PR #16, opened 2018-11-08
Add attachment parameter to payments api

bodrych:
It's very small, but very necessary change for me


bodrych:
Will trailing commas be useful?

PR #63, opened 2019-06-15
Fix API method getKEK bug

VasiliyIsaichkin:


finico:
@VasiliyIsaichkin thank you

PR #101, opened 2021-05-05
Fix Failed to construct URL

### LINKED ISSUE #100: [BUG] Failed to construct URL
**Describe the bug**

When interacting with a canvas, I am seeing that error in the console + the code injected can break some canvas application I am interacting with

**To Reproduce**
1. Go to 'http://draw.io/'
2. Add a svg image element to the canvas
3. Double click the element on the canvas

**Expected behavior**

I can see that this is coming from the `checkForPaymentApiLink` function which is checking the `href` property which exists in `a` element but also in the `SVGAnimatedString` element.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: OSX
 - Browser/JVM chrome
 - Version 1.2.1

**Additional context**

oktapodia:
Fix #100

### DIFF
--- a/src/inpage.js
+++ b/src/inpage.js
@@ -126,6 +126,10 @@ function checkForPaymentApiLink(e) {
     const check = (node) => {
         const href = node.href;
 
+        if(node.tagName.toLowerCase() !== 'a') {
+            return false
+        }
+
         if (!node.href) {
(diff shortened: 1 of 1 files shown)


isakovanton:
Fixed in #111

PR #138, opened 2021-11-05
Editable fee for Transfer / Invoke Script transactions

smelnikov:


src/ui/components/ui/select/Select.tsx
```diff
@@ -120,7 +120,7 @@ interface IProps<T> extends React.ComponentProps<'div'> {
   selectList: Array<TSelectItem<T>>;
   description?: TText;
   selected?: string | number;
-  onSelectItem?: (id: string | number) => void;
+  onSelectItem?: (id: string | number, value?: T) => void;
```
faergeek, line 123: Why is `value` optional? Looks like it's always passed on call.

src/ui/actions/messages.ts
```diff
@@ -19,3 +20,9 @@ export const rejectForever = messageId => ({
   type: ACTION.REJECT_FOREVER,
   payload: { messageId, forever: true },
 });
+
+export const updateTransactionFee = (messageId: string, fee: IMoneyLike) => ({
+  type: ACTION.UPDATE_TRANSACTION_FEE,
+  payload: fee,
+  meta: { messageId },
```
faergeek, line 27: I guess `messageId` should be put into `payload` instead of `meta`

faergeek (approved):

src/controllers/BalanceController.js
```diff
@@ -53,32 +53,45 @@ export class BalanceController {
...
+              available: available.toString(),
+              leasedOut: leasedOut.toString(),
+              network: currentNetwork,
+              assets: Object.fromEntries(
+                assetBalances.balances.map(info => [
+                  info.assetId,
+                  {
+                    minSponsoredAssetFee:
+                      info.minSponsoredAssetFee &&
+                      new BigNumber(info.minSponsoredAssetFee).toString(),
```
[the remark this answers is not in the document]
smelnikov: We need to do a deeper research on `BigNumber` in our project.

faergeek (approved):

PR #143, opened 2021-11-08
Account import metrics

smelnikov:


faergeek (approved):
In general I can't imagine why calling statistics tracking through redux can be useful. May be it makes sense to just call the background methods directly instead?

PR #148, opened 2021-11-12
Redesign account switching

faergeek:


smelnikov (approved):

smelnikov (approved):

msmolyakov (approved):
It would be great to display toast about successful import of accounts, and then it can be merged

PR #150, opened 2021-11-15
Assets balances

smelnikov:


src/ui/components/pages/assets/assetInfo.tsx
```diff
@@ -0,0 +1,149 @@
...
+interface Props {
+  asset: AssetDetail;
+  onCopy: () => void;
+  onClose: () => void;
+}
+
+export function AssetInfo({ asset, onCopy, onClose }: Props) {
+  const selectedAccount = useAppSelector(state => state.selectedAccount);
+
+  const network = selectedAccount && selectedAccount.networkCode;
```
faergeek, line 20: I doubt `selectedAccount` can be falsy here. Also, you can get `networkCode` right in the selector. And I would name it `networkCode` as well. I guess it's a good idea to keep `network` for something like "mainnet" or "testnet" and `networkCode` for "W" or "T". So, something like this should work:

```ts
const networkCode = useAppSelector(state => state.selectedAccount.networkCode);
```

src/ui/components/pages/assets/assetInfo.tsx
```diff
@@ -0,0 +1,149 @@
...
+        </div>
+
+        <div className="margin-main">
+          <div className="input-title basic500 tag1">
+            <Trans i18nKey="assetInfo.displayName" />
+          </div>
+          <div className="tag1">{asset.displayName}</div>
+        </div>
+
+        {!!asset.ticker && (
```
faergeek, line 54: It's a string, so there's no reason to convert it into a boolean here, right?

src/ui/components/pages/assets/assetInfo.tsx
```diff
@@ -0,0 +1,149 @@
...
+        </div>
+
+        <div className="margin-main">
+          <div className="input-title basic500 tag1">
+            <Trans i18nKey="assetInfo.type" />
+          </div>
+          <div className="tag1">
+            <Trans
+              i18nKey={
+                !!asset.reissuable
```
faergeek, line 86: No need for `!!`

src/ui/components/pages/assets/assetInfo.tsx
```diff
@@ -0,0 +1,149 @@
...
+                  ? 'assetInfo.reissuable'
+                  : 'assetInfo.notReissuable'
+              }
+            />
+            ,
+            <Trans
+              i18nKey={
+                !!asset.hasScript
```
faergeek, line 94: No need for `!!`

src/ui/components/pages/assets/assetLogo.tsx
```diff
@@ -0,0 +1,62 @@
...
+const Sponsored = (
+  <svg viewBox="0 0 10 10" className={styles.assetSubIconSvg}>
+    <path
+      fill="currentColor"
+      d="M7.649 2.351a.714.714 0 0 1 0 1.012L3.363 7.649A.715.715 0 0 1 2.35 6.637l4.286-4.286a.714.714 0 0 1 1.012 0zm-5.22 1.506a1.429 1.429 0 1 1 0-2.857 1.429 1.429 0 0 1 0 2.857zM7.57 9a1.429 1.429 0 1 1 0-2.857 1.429 1.429 0 0 1 0 2.857z"
+    />
+  </svg>
+);
+
+interface Props extends React.HTMLProps<HTMLImageElement> {
```
faergeek, line 25: I think extending `React.HTMLProps<HTMLImageElement>` is confusing here. You don't have a rest/spread for other props and you don't always render an `img`. So why would you receive all the props `img` tag would receive here? Better just add `className` prop directly instead of extending.

src/ui/components/pages/assets/assetLogo.tsx
```diff
@@ -0,0 +1,62 @@
...
+    backgroundColor: new ColorHash().hex(assetId),
+  };
+  if (!logo) {
+    return (
+      <div className={cn(styles.assetLogo, className)} style={style}>
+        <div>{name && name[0].toUpperCase()}</div>
+        {(hasSponsorship || hasScript) && (
+          <div className={styles.assetSubIconContainer}>
+            <div className={styles.assetSubIcon}>
+              {hasSponsorship ? Sponsored : Scripted}
```
faergeek, line 51: It's a bit strange to use uppercase for `Sponsored` and `Scripted`. They're not components nor classes. I'd propose either making them components or rename constants with lowercase. And personally I would just inline their contents right here.

src/ui/components/ui/balance/Balance.tsx
```diff
@@ -26,19 +26,25 @@ const BalanceComponent = ({
...
 
+  React.useEffect(() => {
+    if (!assets[assetId]) {
+      getAsset(assetId);
+    }
+  }, []);
+
   switch (true) {
+    case !assets[assetId]:
+      return <Loading>{children}</Loading>;
```
faergeek, line 42: May be merge these two `case`s into one? The `return` statement is the same.

src/ui/services/Background.ts
```diff
@@ -229,6 +231,10 @@ class Background {
     }
   }
 
+  async nftInfo(): Promise<Asset[]> {
```
faergeek, line 234: Looks like it actually returns plain objects instead of `Asset` instances. Or do I miss something?

src/ui/styles/colors.styl
```diff
@@ -1,4 +1,5 @@
 $black = #000;
+$deepGray = #292f3c;
```
faergeek, line 2: I wouldn't bother adding more color variables here. I guess these are from mockups mainly. And AFAIK we're going to get rid of stylus anyway.

faergeek (approved):

faergeek (approved):

msmolyakov (approved):

PR #156, opened 2021-11-23
Transaction history

smelnikov:


faergeek (approved):

src/controllers/TxInfoController.js
```diff
@@ -29,4 +27,49 @@ export class TxInfoController {
...
+    switch (resp.status) {
+      case 200:
+        return resp
+          .text()
+          .then(text =>
+            JSON.parse(
+              text.replace(/(".+?"[ \t\n]*:[ \t\n]*)(\d{15,})/gm, '$1"$2"')
+            )
+          )
+          .then(arr => arr.length && arr[0]);
```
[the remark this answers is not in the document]
smelnikov: > why there's a nested array returned
According to the node developer, this is a historical bug, the fixing of which will lead to a violation of backward compatibility.

src/controllers/TxInfoController.js
```diff
@@ -29,4 +27,49 @@ export class TxInfoController {
...
+
+    let resp = await fetch(url);
+
+    switch (resp.status) {
+      case 200:
+        return resp
+          .text()
+          .then(text =>
+            JSON.parse(
+              text.replace(/(".+?"[ \t\n]*:[ \t\n]*)(\d{15,})/gm, '$1"$2"')
```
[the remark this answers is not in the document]
smelnikov: Yes, it seems like a good idea

src/controllers/TxInfoController.js
```diff
@@ -45,7 +45,10 @@ export class TxInfoController {
               text.replace(/(".+?"[ \t\n]*:[ \t\n]*)(\d{15,})/gm, '$1"$2"')
             )
           )
-          .then(arr => arr.length && arr[0]);
+          .then(arr => (Array.isArray(arr) && arr[0]) || []);
```
faergeek: I doubt this check is needed. Wouldn't this be enough?

```js
.then(arr => arr[0])
```

src/controllers/TxInfoController.js
```diff
@@ -29,4 +27,44 @@ export class TxInfoController {
...
+      this.getNode()
+    ).toString();
+
+    let resp = await fetch(url, {
+      headers: { accept: 'application/json; large-significand-format=string' },
+    });
+
+    switch (resp.status) {
+      case 200:
+        return resp.json().then(arr => (Array.isArray(arr) && arr[0]) || []);
```
faergeek, line 43: I doubt this `Array.isArray(...)` check is needed. Wouldn't this be enough?

```js
.then(arr => arr[0])
```

smelnikov: I missed that node returns `[ [ ] ]` if there are no records. Of course, that will be enough.

PR #161, opened 2021-12-02
Assets filters

smelnikov:


src/ui/components/pages/Assets.tsx
```diff
@@ -3,72 +3,99 @@ import cn from 'classnames';
...
+          }}
+        >
+          <svg
+            width="14"
+            height="14"
+            viewBox="0 0 18 18"
+            fill="none"
+            xmlns="http://www.w3.org/2000/svg"
+          >
+            <rect
```
faergeek, line 61: It's an empty rect used to size icons, which can be removed. And by the way, `xmlns` is not needed on `svg` when it's not a separate svg file.

src/ui/components/pages/assets/assetItem.styl
```diff
@@ -11,15 +11,77 @@
   position: relative;
   z-index: 1;
   margin-bottom: 8px;
-  cursor: pointer;
+  cursor: default;
```
faergeek, line 14: It's better to remove `cursor` prop here. With `cursor: default` it looks like you can't select asset name or balance, because it always looks like an arrow. Without this prop cursor will change to a vertical line when it's above the text.

src/ui/components/pages/assets/assetItem.styl
```diff
@@ -11,15 +11,77 @@
...
-.assetName
+.assetTitle
+  display: flex;
+  justify-content: flex-start;
   height: 24px;
-  max-width: 220px;
+  max-width: 180px;
+
+.assetName
+  padding-right: 4px;
```
faergeek, line 28: May be use the `gap` property on `.assetTitle` instead? It creates a gap between child elements.

src/ui/components/pages/assets/assetItem.styl
```diff
@@ -11,15 +11,77 @@
...
   height: 64px;
 
 .assetData
   flex-grow: 1;
   text-align: left;
 
-.assetName
+.assetTitle
+  display: flex;
+  justify-content: flex-start;
```
faergeek, line 23: Isn't this the default value?

src/ui/components/pages/assets/assetItem.tsx
```diff
@@ -28,18 +34,71 @@ export function AssetItem({ balance, assetId, className, onClick }: Props) {
...
+          {<div className={styles.assetName}>{displayName}</div> || <Loader />}
+          {asset?.isSuspicious && (
+            <svg
+              className={styles.assetStatusIcon}
+              width="18"
+              height="18"
+              viewBox="0 0 18 18"
+              fill="none"
+            >
+              <rect width="18" height="18" fill="#D8D8D8" fillOpacity="0.01" />
```
faergeek, line 47: Empty rect, can be removed

src/ui/components/pages/assets/assetItem.tsx
```diff
@@ -28,18 +34,71 @@ export function AssetItem({ balance, assetId, className, onClick }: Props) {
...
+              ? 'assetInfo.favRemoveTooltip'
+              : 'assetInfo.favAddTooltip'
+          }
+        />
+      </div>
+
+      <button
+        className={cn(styles.infoButton, 'showTooltip')}
+        type="button"
+        onClick={() => onClick(assetId)}
```
faergeek, line 92: May be rename `onClick` prop to `onInfoClick`?

src/ui/components/pages/assets/tabs/helpers.ts
```diff
@@ -0,0 +1,131 @@
...
+import {
+  AssetFilters,
+  NftFilters,
+  TxHistoryFilters,
+} from '../../../../reducers/updateState';
+
+function useFilter<T, K extends keyof T>(
+  name: string,
+  field: K
+): [T[K], (value: any) => void] {
```
faergeek, line 13: Doesn't `T[K]` works for an argument also?

src/ui/components/pages/assets/tabs/tabAssets.tsx
```diff
@@ -0,0 +1,121 @@
...
+        (!onlyMyAssets || assets[assetId]?.issuer === address) &&
+        (!assetTerm ||
+          assetId === assetTerm ||
+          icontains(assets[assetId]?.displayName, assetTerm))
+    )
+    .sort(
+      ([a], [b]) =>
+        (assets[a] &&
+          assets[b] &&
+          +!!assets[b].isFavorite - +!!assets[a].isFavorite) ||
```
faergeek, line 42: I guess parenthesis are grouped incorrectly here. Opening paren should be on this line, not before the first condition, where you check for assets presence. Otherwise, if there's no assets for these ids, the following conditions (e.g. `isSuspicious`) are checked, reading props from `undefined`. I got an error when I open the popup for the first time after this change, before assets are loaded.

faergeek (changes requested):

src/ui/components/pages/assets/tabs/tabAssets.tsx
smelnikov: Sure, already fixed this

src/ui/components/pages/assets/assetItem.styl
smelnikov: It looks like we don't need gap or padding here.

smelnikov: Oh, I forgot that the `.assetTitle` has icons, so the ` gap` will be very useful here.

src/background.js
```diff
@@ -338,7 +341,7 @@ class BackgroundService extends EventEmitter {
       WalletController: this.walletController.store,
       NetworkController: this.networkController.store,
       MessageController: this.messageController.store,
-      BalanceController: this.balanceController.store,
+      CurrentAccountController: this.currentAccountController.store,
```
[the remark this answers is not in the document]
smelnikov: I think it's okay, the size of the balance controller is too small to affect performance.

src/controllers/AssetInfoController.js
```diff
@@ -112,37 +119,109 @@ export class AssetInfoController {
...
+    this.store.updateState({ assets });
+  }
+
+  /**
+   * Force-updates storage asset info by assetId list
+   * @param {string} forAddress
+   * @param {Array<string>} assetIds
+   * @returns {Promise<void>}
+   */
+  async updateAssets(forAddress, assetIds) {
```
faergeek, line 141: `forAddress` isn't needed here, right?

src/ui/services/Background.ts
```diff
@@ -209,19 +203,9 @@ class Background {
     }
   }
 
-  async nftInfo(address: string, limit: number = 1000): Promise<AssetDetail[]> {
+  async assetFavorite(assetId: string): Promise<AssetDetail> {
```
faergeek, line 206: The return type should be `Promise<void>` now, since it's not returned from background.js and controller

src/controllers/AssetInfoController.js
faergeek: Removed it here https://github.com/wavesplatform/waves-keeper/pull/166/commits/f6b02ba7e71b8a9cf95c72c400510149ab41f084

smelnikov: > `forAddress` isn't needed here, right?

Yeah, it is not necessary. Picked your commit.

src/ui/components/pages/assets/helpers.ts
```diff
@@ -5,9 +5,39 @@ const explorerUrls = new Map([
...
+
+export const colors = {
+  white: '#FFFFFF',
+  basic200: '#DAE1E9',
+  basic500: '#9BA6B2',
+  submit200: '#BACAF4',
+  submit400: '#1F5AF6',
+  out: '#FFAF00',
+  in: '#81C926',
+};
```
faergeek, line 40: Not sure about `out` and `in` as being generic enough, but other colors are the same as in `colors.styl`. May be use css custom properties for them? With something like this in `app.styl`:

```css
:root {
  --color-white: #fff;
  --color-basic200: #dae1e9;
  /* etc. */
}
```

And then it can be used as `var(--color-basic200)` in css or for coloring svgs.

faergeek (approved):

src/controllers/AssetInfoController.js
```diff
@@ -13,10 +14,16 @@ const WAVES = {
...
 };
+const SUSPICIOUS_LIST_URL =
+  'https://raw.githubusercontent.com/wavesplatform/waves-community/master/Scam%20tokens%20according%20to%20the%20opinion%20of%20Waves%20Community.csv';
+const MAX_AGE = 60 * 60 * 1000;
 
 export class AssetInfoController {
   constructor(options = {}) {
     const defaults = {
+      lastUpdated: {
+        suspiciousAssets: 0,
```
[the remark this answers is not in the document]
smelnikov: Yeah, we can revert this.

src/controllers/AssetInfoController.js
```diff
@@ -32,18 +39,39 @@ export class AssetInfoController {
         },
       },
     };
+    this.suspiciousAssets = undefined;
     this.getNode = options.getNode;
     this.getNetwork = options.getNetwork;
     this.store = new ObservableStore(
-      Object.assign({}, defaults, options.initState)
+      R.mergeDeepRight(defaults, options.initState || {})
```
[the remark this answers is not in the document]
smelnikov: Yeah, we can revert this.

src/controllers/CurrentAccountController.js
```diff
@@ -0,0 +1,191 @@
...
+import { BigNumber } from '@waves/bignumber';
+
+export const MAX_TX_HISTORY_ITEMS = 101;
+const MAX_NFT_ITEMS = 1000;
+
+export class CurrentAccountController {
+  constructor(options = {}) {
+    const defaults = {
+      balances: {},
+      pollInterval: options.pollInterval || 10000,
```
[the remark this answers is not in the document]
smelnikov: Good idea.

src/controllers/CurrentAccountController.js
```diff
@@ -0,0 +1,191 @@
...
+
+          return [
+            address,
+            {
+              available: available.toString(),
+              leasedOut: leasedOut.toString(),
+              network: currentNetwork,
+              ...(isActiveAddress
+                ? {
+                    aliases: aliases || [],
```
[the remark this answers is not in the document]
smelnikov: Nope, same as MyAssets.balance and myNfts

src/ui/components/pages/Messages.tsx
```diff
@@ -75,29 +74,16 @@ class MessagesComponent extends React.Component {
...
-    ].includes(sourceSignData.type)
-      ? Object.keys(sponsoredBalance)
-      : [];
+
     const needGetAssets = new Set(
-      Object.keys(parsedData.assets)
-        .concat(sponsoredAssets)
-        .filter(id => !Object.keys(assets).includes(id))
+      Object.keys(parsedData.assets).filter(
+        id => !Object.keys(assets).includes(id)
```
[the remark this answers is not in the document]
smelnikov: Sure

PR #205, opened 2022-01-27
Show asset tickers instead of names

faergeek:


src/controllers/AssetInfoController.js
```diff
@@ -62,9 +149,32 @@ export class AssetInfoController {
     this.store = new ObservableStore(
       Object.assign({}, defaults, options.initState)
     );
+    this.addTickersForExistingAssets();
```
smelnikov, line 152: It looks like we don't need to update the tickers every time the extension is loaded.
Maybe move this logic to `background.js` for the extension update event handler?!

https://github.com/wavesplatform/waves-keeper/blob/aa1e91a6e17b7793673185686a370e93dc79118c/src/background.js#L98

smelnikov (approved):

PR #210, opened 2022-02-07
Fix notification window size

faergeek:


src/ui.tsx
```diff
@@ -102,6 +102,13 @@ async function startUi() {
     await background.closeNotificationWindow();
   }
 
+  if (window.location.href.split('/').reverse()[0] === 'notification.html') {
```
smelnikov, line 105: Why not just use `window.location.pathname === '/notification.html'`?

smelnikov (approved):

PR #219, opened 2022-02-15
Import from encoded seed and private key

faergeek:


src/wallets/encodedSeed.ts
```diff
@@ -0,0 +1,88 @@
...
+    throw new Error('Cannot get seed');
+  }
+
+  getEncodedSeed() {
+    return this.data.encodedSeed;
+  }
+
+  getPrivateKey() {
+    return libCrypto.privateKey(libCrypto.base58Decode(this.data.encodedSeed));
+  }
```
smelnikov, line 61: Seems like it should be written like this
```js
getSeed(): string {
  return libCrypto.base58Decode(this.getEncodedSeed())
}

getEncodedSeed() {
  return this.data.encodedSeed;
}

getPrivateKey() {
  return libCrypto.privateKey(this.getSeed());
}
 ```

File: src/wallets/seed.ts [hunk 2]
```diff
@@ -0,0 +1,73 @@
...
+    this._adapter = new SeedAdapter(seed, networkCode);
+  }
+
+  getSeed() {
+    return this.data.seed;
+  }
+
+  getPrivateKey() {
+    return libCrypto.privateKey(this.data.seed);
+  }
```
smelnikov, line 46: Similar to previous
```js
getPrivateKey() {
  return libCrypto.privateKey(this.getSeed());
}
```

smelnikov (approved):
GJ!

src/wallets/encodedSeed.ts
faergeek: `base58Decode` returns `Uint8Array` and encoded seed can contain arbitrary binary data. That's why `getSeed` throws for now. Not sure if we should check the contents and decide to throw or not. In UI we only allow to copy encoded seed back

PR #230, opened 2022-03-07
Add ledger support

faergeek:


src/ledger/service.ts
```diff
@@ -0,0 +1,190 @@
...
+            err.message
+          )
+        ) {
+          this._connectionRetryIsNeeded = true;
+        } else {
+          console.error('NO MATCH FOR ERROR', err);
+        }
+      } else {
+        console.error('NON-ERROR THROWN', err);
+      }
```
smelnikov, line 87: Should Sentry catch these errors or is it just to be ignored?

smelnikov (commented):
Great job!

smelnikov (approved):

src/ledger/service.ts
faergeek: No, now sentry won't catch them, but I'll fix it

PR #236, opened 2022-03-17
Open new tab on account creation or import

smelnikov:


src/ui/components/pages/NewWalletName.tsx
```diff
@@ -1,174 +1,148 @@
-import * as styles from './styles/newaccountname.styl';
+import * as styles from 'ui/components/pages/newWalletName.module.css';
```
faergeek, line 1: May be use relative import?))

src/ui/components/pages/NewWalletName.tsx
```diff
@@ -1,174 +1,148 @@
...
+          <Input
+            data-testid="newAccountNameInput"
+            className="margin1"
+            onChange={e => setAccountName(e.target.value)}
+            value={accountName}
+            maxLength="32"
+            disabled={existedAccount}
+            autoFocus
+            error={error}
+            onFocus={() => setTouched(true)}
```
faergeek, line 99: Was it changed from `onBlur` to `onFocus` intentionally? Because of `autoFocus`, `onFocus` is called right away. So in that case it doesn't make sense to think about `touched` state at all, it will be `false` only for short period of time after mount.

src/ui/components/pages/NewWalletName.tsx
```diff
@@ -1,174 +1,148 @@
...
+  const dispatch = useAppDispatch();
+
+  const account = useAppSelector(state => state.localState.newAccount);
+  const accounts = useAppSelector(state => state.accounts);
+  const [accountName, setAccountName] = React.useState<string>('');
+  const [pending, setPending] = React.useState<boolean>(false);
+  const [error, setError] = React.useState<string>('');
+  const [touched, setTouched] = React.useState<boolean>(false);
+
+  const existedAccount = accounts.find(
```
faergeek, line 27: I guess it's not needed to search for an existing account here, because switching is done on previous step. Or is it needed in some cases?

src/ui/components/pages/otherAccounts.tsx
```diff
@@ -106,7 +108,13 @@ export function OtherAccountsPage({ setTab }: Props) {
             className={styles.addAccountButton}
             data-testid="addAccountButton"
             type="button"
-            onClick={() => setTab(PAGES.IMPORT_FROM_ASSETS)}
+            onClick={() => {
+              background.showTab(
+                `${window.location.origin}/accounts.html`,
+                'accounts'
+              );
+              dispatch(resetTab(PAGES.ROOT));
```
faergeek, line 116: Is it necessary to change screen after opening a tab?

faergeek (approved):

src/ui/components/pages/otherAccounts.tsx
smelnikov: Yeah, this is necessary so that the tests pass and do not violate the user flow

src/ui/components/pages/NewWalletName.tsx
smelnikov: This prevents some bugs with multiple accounts tabs.

src/ui/components/pages/NewWalletName.tsx
smelnikov: Sure

src/lib/tabsManager.ts
```diff
@@ -0,0 +1,41 @@
...
+  async getOrCreate(url: string, key: string) {
+    const currentTab = this._tabs[key];
+    const tabProps = { active: true } as Tab;
+    if (url != currentTab?.url) {
+      tabProps.url = url;
+    }
+
+    return new Promise((resolve, reject) => {
+      browser.tabs.create({ url: 'about:blank' }, tab => {
+        browser.tabs.remove(tab.id, () => {
```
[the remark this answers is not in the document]
smelnikov: This was a trick to close the popup in some browsers, but it doesn't work in Firefox.
So now we will close the popup window programmatically.

faergeek (approved):

PR #245, opened 2022-03-29
Fix default button type

ehassed3:


smelnikov (commented):
It seems that the component has become more complex.
Maybe we should use the `<button>` element, e.g. 
`<button className={buttonStyles.primary} type="submit"/>`
or
`<button className={buttonStyles.default} type="button"/>`

ehassed3:
>

ehassed3:
I agree that the component has become more complex because of the exported types. But the logic of the component can be expanded in the future, I left it.

faergeek (approved):
Great job! 👍 

What about making `button` the default value for `type` prop? In html it defaults to `submit` and it often causes an unwanted behavior of submitting surrounding form)

PR #246, opened 2022-03-29
Add set script transaction cancel

ehassed3:


src/ui/components/transactions/SetScript/SetScriptCard.tsx
```diff
@@ -3,8 +3,8 @@ import * as React from 'react';
 import { Trans } from 'react-i18next';
 import { TxIcon } from '../BaseTransaction';
```
faergeek, line 4: This import is not used anymore, can be removed

src/ui/components/transactions/SetScript/SetScriptCardHeader.module.css
```diff
@@ -0,0 +1,21 @@
...
+}
+
+.icon {
+  height: 48px;
+  margin-right: 10px;
+  width: 48px;
+}
+
+.data {
+  color: #9ba6b2;
```
faergeek, line 13: You can use `var(--color-basic500)` here instead. Colors are declared in `app.styl`

src/ui/components/transactions/SetScript/SetScriptCardHeader.tsx
```diff
@@ -0,0 +1,34 @@
+import * as React from 'react';
+import * as styles from './SetScriptCardHeader.module.css';
+import { useTranslation } from 'react-i18next';
+
+import { TxIcon } from '../BaseTransaction';
+
+interface IProps {
```
faergeek, line 7: No need for `I` prefix here, just `Props` would be enough

src/ui/components/transactions/SetScript/SetScriptCardHeader.tsx
```diff
@@ -0,0 +1,34 @@
...
+  script?: string;
+}
+
+const SetScriptCardHeader: React.FC<IProps> = ({ script }) => {
```
faergeek, line 11: Is there any reason to use `React.FC` type? Function declaration like this would be enough:

```typescript
function SetScriptCardHeader({ script }: Props) {
```

src/ui/components/transactions/SetScript/SetScriptCardHeader.tsx
```diff
@@ -0,0 +1,34 @@
...
+    <div className={styles.header}>
+      <TxIcon className={styles.icon} txType="set-script-cancel" />
+      <h1 className={styles.title}>
+        {t('transactions.setScriptTransactionCancel')}
+      </h1>
+    </div>
+  );
+};
+
+export default SetScriptCardHeader;
```
faergeek, line 34: Named exports are preferred. Right at the point of declaration, like this:

```typescript
export function SetScriptCardHeader(...) {
```

faergeek (approved):
This is not reflected in the task, but I guess "Remove Account Script" would fit better than "Cancel Account Script". @msmolyakov do you agree?

msmolyakov:
> This is not reflected in the task, but I guess "Remove Account Script" would fit better than "Cancel Account Script". @msmolyakov do you agree?

Agree, sounds better. But it'is important to rename it in all places: transactions history, confirmation screen, etc.

faergeek (approved):

PR #251, opened 2022-04-04
Add usd amount

ehassed3:


src/controllers/AssetInfoController.js
```diff
@@ -17,6 +17,19 @@ const SUSPICIOUS_LIST_URL =
...
+const MARKETDATA_POLL_INTERVAL = 10 * 60 * 1000;
+
+const stablecoinAssetTickers = {
+  '2thtesXvnVMcCnih9iZbJL3d2NQZMfzENJo8YFj6r5jU': 'UST',
+  '34N9YcEETLWn93qYQ64EsP1x89tSruJU44RrEMSXXEPJ': 'USDT',
+  '6XtHjpXbs9RRJP2Sr9GUyVqzACcby9TkThHXnjVC5CDJ': 'USDC',
+  '8DLiYZjo3UUaRBTHU7Ayoqg4ihwb6YH1AfXrrhdjQ7K1': 'BUSD',
+  '8zUYbdB8Q6mDhpcXYv52ji8ycfj4SDX4gJXS7YY3dA4R': 'DAI',
+  DG2xFkPdDwKUoBkzGAhQtLpSGzfXLiCYPEzeKH2Ad24p: 'USDN',
+};
```
smelnikov, line 31: it is enough to use a `Set` here and check the occurrence with `Set.prototype.has()`

src/ui/components/accounts/activeAccountCard.tsx
```diff
@@ -23,7 +21,7 @@ interface Props {
 
 export function ActiveAccountCard({
   account,
-  balance,
+  amount,
```
smelnikov, line 24: Might use a more appropriate name for this prop, e.g. `amountInUsd`?!

src/ui/components/pages/assets/assetItem.tsx
```diff
@@ -86,7 +86,17 @@ export function AssetItem({
         </div>
 
         <div className={styles.balance}>
-          <Balance isShortFormat={false} split={true} balance={balance} />
+          {!isLoading ? (
```
smelnikov, line 89: Is it necessary? Asset loading check is done inside the `Balance` component

src/ui/components/transactions/ScriptInvocation/ScriptInvocationCard.tsx
```diff
@@ -117,7 +96,19 @@ export class ScriptInvocationCard extends React.PureComponent<IProps> {
                   className={styles.expandableList}
                   showExpand={!collapsed}
                 >
-                  <AmountTable amounts={amounts} />
+                  <>
```
smelnikov, line 99: You can skip wrapping in `<Fragment>` here

src/ui/components/ui/UsdAmount/index.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+import { AssetDetail } from 'ui/services/Background';
+
+interface Props {
+  asset?: AssetDetail;
+  amount: number;
+  className?: string;
+}
+
+export const UsdAmount = ({ asset, amount, className }: Props) =>
```
smelnikov, line 10: Better use `function` to define this component

src/ui/components/ui/balance/Balance.module.css
```diff
@@ -0,0 +1,10 @@
+.usdAmount {
+  color: #9ba6b2;
```
smelnikov, line 2: We have defined css color variables at `app.styl`, so replace it with `color: var(--color-basic500)`

src/ui/services/Background.ts
```diff
@@ -558,10 +558,11 @@ export enum WalletTypes {
   KeystoreWx = 'keystore_wx',
 }
 
-export interface AssetDetail extends IAssetInfo {
+export interface AssetDetail extends Asset {
```
smelnikov, line 561: This interface defines the structure of plain objects, not instances of classes, which is why `IAssetInfo` was previously used.

src/ui/components/accounts/activeAccountCard.tsx
ehassed3: It seemed to me that the `amount` is enough, because usd are the only amount

src/ui/components/pages/assets/assetItem.tsx
ehassed3: This worked incorrectly — when there is no list of assets, a `PLACEHOLDER` is passed, where the array index is used as the `assetId`, which caused side work inside `Balance` component

src/ui/components/transactions/ScriptInvocation/ScriptInvocationCard.tsx
ehassed3: `PlateCollapsable` component does not allow me to do this

src/ui/services/Background.ts
smelnikov: `AssetDetails`  determines only fields of plain object. 
That means in functions that returns this type is not instances of `Asset` class (cuz they don't have methods implementation)

src/ui/components/pages/assets/assetItem.tsx
smelnikov: https://github.com/wavesplatform/waves-keeper/blob/KEEP-277/add-usd-amount/src/ui/components/ui/balance/Balance.tsx#L56:L57

In this case, we will see loader.
This logic does the same thing as `isLoading`.
When we hit `assets[1]`, undefined is returned and the loader is displayed.

faergeek: `assetId` isn't passed into `Balance` component here, so the prop defaults to `'WAVES'`. And for `'WAVES'` there's no request sent to the node, because it's info is hard-coded inside `AssetInfoController`.

faergeek: So the only side effect of this is that the request is sent from popup to the background and it stops there. But may be it makes sense to add a loader here directly to avoid all of that

smelnikov: > So the only side effect of this is that the request is sent from popup to the background and it stops there. But may be it makes sense to add a loader here directly to avoid all of that

I think this is right solution.

smelnikov:
You should also fix assertions in the test

https://github.com/wavesplatform/waves-keeper/blob/98a963a95c62b087ae66705390791bf6023ae865/test/Others.ui.ts#L143:L145

https://github.com/wavesplatform/waves-keeper/blob/98a963a95c62b087ae66705390791bf6023ae865/test/Others.ui.ts#L188:L190

smelnikov (changes requested):
You should also fix assertions in the test

https://github.com/wavesplatform/waves-keeper/blob/98a963a95c62b087ae66705390791bf6023ae865/test/Others.ui.ts#L143:L145

https://github.com/wavesplatform/waves-keeper/blob/98a963a95c62b087ae66705390791bf6023ae865/test/Others.ui.ts#L188:L190

src/controllers/AssetInfoController.js
```diff
@@ -142,6 +155,7 @@ export class AssetInfoController {
         },
       },
     };
+    this.usdPriceAssets = undefined;
```
faergeek, line 158: I guess naming it `usdPrices` makes more sense. Right now it makes me think that it contains assets representing usd prices for us. But that doesn't seem to be true. Plus the method `updateUsdPriceAssets` better be named `updateUsdPrices`. Or do I miss something?

src/ui/components/accounts/activeAccountCard.tsx
```diff
@@ -42,7 +40,13 @@ export function ActiveAccountCard({
             {account.name}
           </div>
 
-          <Balance balance={balance} isShortFormat={false} showAsset split />
+          {amountInUsd !== null ? (
+            <p className={styles.accountAmount}>{`$${
+              Math.ceil(+amountInUsd * 100) / 100
```
faergeek, line 45: It's better to use `BigNumber` class for things like this to preserve precision. And so `amountInUsd` prop should accept `BigNumber | null`. It has `toFixed` and `toFormat` methods.

src/ui/components/pages/Assets.tsx
```diff
@@ -72,20 +72,26 @@ export function Assets({ setTab }: Props) {
...
-  if (assetInfo) {
-    const asset = new Asset(assetInfo);
-    wavesBalance = new Money(balances[address]?.available || 0, asset);
-  }
+  const amountInUsd = balances[address]?.assets
+    ? Object.entries(balances[address].assets).reduce(
+        (acc, [id, { balance }]) => {
+          if (assets[id]?.usdPrice) {
+            const tokens = new Money(balance, new Asset(assets[id])).toTokens();
+            acc += +assets[id].usdPrice * +tokens;
```
faergeek, line 80: If you use `getTokens` instead of `toTokens`, you'll get `BigNumber`, and then `acc` calculation would look like this:

```typescript
acc.plus(new BigNumber(assets[id].usdPrice).mul(tokens));
```

And `reduce` starting value would be `new BigNumber(0)`

src/ui/components/ui/UsdAmount/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+}
+
+export function UsdAmount({ asset, amount, className }: Props) {
+  if (!asset?.usdPrice || asset.usdPrice === '1') {
+    return null;
+  }
+
+  return (
+    <p className={className}>{`≈ $${
+      Math.ceil(+asset.usdPrice * amount * 100) / 100
```
faergeek, line 17: Use `BigNumber` here

faergeek (approved):

faergeek (approved):

PR #252, opened 2022-04-06
Fix reject cancel lease card

ehassed3:


faergeek:
I think it's a good idea to fetch lease data for package as it's done here for a single transaction
https://github.com/wavesplatform/waves-keeper/blob/cab12f91ca598e96a8c886f567964deaba546346/src/controllers/MessageController.js#L748

src/ui/components/transactions/Package/parseTx.ts
```diff
@@ -9,7 +9,7 @@ export const txType = 'transactionPackage';
 export function getTransactionData(item) {
   const tx = { type: item.type, ...(item.data ? item.data : item) };
   const config = getConfigByTransaction({ data: item, type: 'transaction' });
-  return { tx, config };
+  return { tx, config, ...(item.lease ? { lease: item.lease } : {} ) };
```
smelnikov, line 12: Maybe `lease` information should be nested in `tx` since that data related to the transaction?!

ehassed3: `lease` data is not used in `tx` in https://github.com/wavesplatform/waves-keeper/blob/cab12f91ca598e96a8c886f567964deaba546346/src/ui/components/transactions/CancelLease/CancelLeaseCard.tsx#L59 and https://github.com/wavesplatform/waves-keeper/blob/cab12f91ca598e96a8c886f567964deaba546346/src/controllers/MessageController.js#L754. Should I fix it then?

smelnikov: Now I see that this interface matches what is used in the message for a single transaction.
Leave it as is.

ehassed3: Fixed

faergeek (approved):

PR #253, opened 2022-04-08
Rename extension

ehassed3:


src/ui/components/pages/importKeystore/importKeystore.tsx
```diff
@@ -49,6 +49,9 @@ type ExchangeKeystoreAccount = {
   | {
       userType: 'wavesKeeper';
     }
+  | {
+      userType: 'keeperWallet';
```
smelnikov, line 53: `ExchangeKeystoreAccount` defines the wave.exchange import file format, so I don't think this field will be renamed in the future.
So maybe it's not necessary?

faergeek (approved):

src/ui/appConfig.ts
```diff
@@ -1,4 +1,4 @@
-export const WAVESKEEPER_DEBUG = process.env.NODE_ENV !== 'production';
+export const KEEPERWALLET_DEBUG = process.env.NODE_ENV !== 'production';
```
msmolyakov, line 1: Does it make sense to keep the previous variable `WAVESKEEPER_DEBUG` as well?

msmolyakov (changes requested):

src/ui/appConfig.ts
ehassed3: I think it doesn't matter

msmolyakov (approved):

PR #259, opened 2022-04-18
Adapt swap screen for multi-routing

faergeek:


src/ui/components/pages/swap/channelClient.ts
```diff
@@ -95,43 +109,21 @@ export class ExchangeChannelClient {
...
-              vendor,
-            }): ExchangePool => ({
-              dApp: address,
-              estimatedAmount: new BigNumber(String(estimatedAmount)),
-              fromAssetId: source,
-              toAssetId: target,
-              type,
-              vendor,
-            })
+      if (res.exchange.data) {
```
smelnikov, line 112: Here's the correct way to check which field is filled.
```javascript
switch (res.exchage.result){
  case "data":
     // ...
  case "error":
     // ...
}```

File: src/ui/components/pages/swap/channelClient.ts [hunk 2]
```diff
@@ -29,34 +34,43 @@ interface ExchangeRequest extends ExchangeInput {
...
+) => void;
 
-  constructor(
-    type: ExchangeChannelErrorType,
-    code = ExchangeChannelErrorCode.UNEXPECTED_ERROR
-  ) {
-    this.type = type;
-    this.code = code;
+function convertArg(arg: proto.Response.Exchange.Argument): CallArg {
+  if ('integerValue' in arg) {
```
smelnikov, line 46: It is better to check which field is filled.
This should work but i could be wrong :)

```javascript
switch (arg.value){
  case "integerValue":
     // ...
  case "binaryValue":
     // ...
  // ...
}```

Comment (smelnikov) [inline src/ui/components/pages/swap/channelClient.ts:46]: Here's the correct way to check which field is filled.
```javascript
switch (arg.value){
  case "integerValue":
     // ...
  case "binaryValue":
     // ...
  // etc
}```

**Review by smelnikov — APPROVED** [review]

File: src/ui/components/pages/swap/form.tsx [hunk 3]
```diff
@@ -155,16 +173,15 @@ export function SwapForm({
...
+  );
 
   const [channelClient, setChannelClient] =
     React.useState<ExchangeChannelClient | null>(null);
 
   React.useEffect(() => {
-    const client = new ExchangeChannelClient(
-      new URL('/v1', swapChannel).toString()
-    );
+    const client = new ExchangeChannelClient('ws://localhost:8765');
```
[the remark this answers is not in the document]
smelnikov: Also this should be `v2`

smelnikov (approved):

src/vendors/constants.ts
```diff
@@ -0,0 +1,11 @@
+export enum SwapVendor {
```
faergeek, line 1: `vendors` sounds too generic. May be better name folder to `swap`?

src/ui/components/pages/swap/form.tsx
```diff
@@ -255,505 +310,698 @@ export function SwapForm({
...
+    [null, SwapVendor.Keeper]
+  );
+
   React.useEffect(() => {
-    const assetsToUpdate = Array.from(
-      new Set(routeAssets.filter(asset => asset == null))
+    if (!touched) {
+      setSelectedExchangeVendor(profitVendor);
+    }
+  }, [touched, nonProfitVendor, profitVendor, exchangeInfo]);
```
faergeek, line 430: Not sure why `nonProfitVendor` and `exchangeInfo` are needed here

PR #260, opened 2022-04-19
Add price precision conversion for create order

faergeek:


src/transactions/utils.ts
```diff
@@ -551,7 +551,15 @@ export const convertFromSa = {
...
-      price: input.data.price.getCoins(),
+      price: input.data.price
+        .getTokens()
+        .mul(
+          new BigNumber(10).pow(
+            8 +
+              input.data.price.asset.precision -
+              input.data.amount.asset.precision
+          )
+        ),
```
smelnikov, line 562: Based on the specification, the difference in decimals is required for version 1 and 2 
https://docs.waves.tech/en/blockchain/transaction-type/exchange-transaction#json-representation

Probably should also fix these checks in the tests

smelnikov (changes requested):

src/transactions/utils.ts
smelnikov: Oh, looked at the wrong section of the documentation. 
It means everything is right.
https://docs.waves.tech/en/blockchain/order#json-representation

smelnikov (approved):

PR #271, opened 2022-04-26
Migrating to manifest v3

ehassed3:


src/ui/services/Background.ts
```diff
@@ -117,24 +141,23 @@ class Background {
...
       await this.initPromise;
+      await this._connect();
       return await this.background.selectAccount(address, network);
     } catch (err) {
       throw new Error(prepareErrorMessage(err));
     }
   }
 
   async addWallet(data): Promise<Account> {
-    try {
```
faergeek, line 127: Why try/catch was removed here?

src/controllers/RemoteConfigController.js
```diff
@@ -61,6 +63,22 @@ export class RemoteConfigController extends EventEmitter {
...
+        case 'updateConfig':
+          this._getConfig();
+          break;
+        case 'updateIgnoreErrorsConfig':
+          this._getIgnoreErrorsConfig();
+          break;
+        case 'fetchIdentityConfig':
+          this._fetchIdentityConfig();
+          break;
+        default:
```
faergeek, line 78: No need for `default`

src/controllers/CurrentAccountController.js
```diff
@@ -22,13 +24,20 @@ export class CurrentAccountController {
...
+        this.updateBalances();
+      }
+    });
     this.restartPolling();
   }
 
   restartPolling() {
-    clearInterval(this.poller);
-    this.poller = setInterval(this.updateBalances.bind(this), POLL_INTERVAL);
+    extension.alarms.clear('updateBalances');
```
faergeek, line 37: There's no need to clear alarm before creation, because alarm with the same name will be replaced, according to docs:

https://developer.chrome.com/docs/extensions/reference/alarms/#method-create
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/alarms/create

There are also other places where you do it.

src/controllers/VaultController.ts
```diff
@@ -19,9 +20,10 @@ export class VaultController {
   constructor(opts: Options) {
     this.store = new ObservableStore(
       Object.assign({}, { locked: null, initialized: null }, opts.initState, {
-        locked: true,
+        locked: !opts.password,
       })
     );
+    this.password = opts.password;
```
faergeek, line 26: It seems `this.password` isn't read anywhere. @smelnikov any idea why it's here?

faergeek (approved):

src/lib/extension.js
```diff
@@ -0,0 +1,79 @@
+const apis = [
```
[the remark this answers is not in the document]
ehassed3: Yes

src/controllers/VaultController.ts
smelnikov: Yeah. Looks like an outdated code snippet.
The original idea was for wallet/identity controllers to look at this field to unlock their encrypted store values.
This was later resolved with dependency injection of these controllers.

```javascript
unlock(password) {
    this.wallet.unlock(password);
    this.identity.unlock(password);

    this.password = password;
    this.locked = false;
  }
  ```

**Review by smelnikov — CHANGES_REQUESTED** [review]

File: src/controllers/IdentityController.ts [hunk 6]
```diff
@@ -83,17 +51,27 @@ class IdentityStorage extends ObservableStore implements ICognitoStorage {
...
+
+    this.memo = initSession.memo;
+
+    this.password = initSession.password;
+    this._setSession = setSession;
   }
 
   lock() {
-    this.password = undefined;
+    this._setPassword(null);
```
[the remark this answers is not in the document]
ehassed3: No, this cannot be done because the manifest v2 (firefox) does not support the storage session.

smelnikov (approved):

src/controllers/CurrentAccountController.js
```diff
@@ -1,9 +1,10 @@
+import { extension } from 'lib/extension';
 import ObservableStore from 'obs-store';
 import { BigNumber } from '@waves/bignumber';
 
 export const MAX_TX_HISTORY_ITEMS = 101;
 const MAX_NFT_ITEMS = 1000;
-const POLL_INTERVAL = 10000;
+const PERIOD_IN_SECONDS = 10000;
```
faergeek, line 7: Guess this should be 10 instead of 10000

PR #279, opened 2022-05-04
Fix flickering test failures

ehassed3:


test/AccountCreation.ui.ts
```diff
@@ -475,13 +443,16 @@ describe('Account creation', function () {
             await accountNameInput.sendKeys(ACCOUNTS.ANY);
             await continueBtn.click();
 
-            const importSuccessForm = await this.driver.wait(
-              until.elementLocated(By.css('[data-testid="importSuccessForm"]')),
-              this.wait
-            );
-            expect(importSuccessForm).not.to.be.throw;
+            expect(
```
faergeek, line 446: Not sure what this `expect` actually checks. @smelnikov ? As I understand this should check if provided function throws on invocation, but we don't have a function here

https://www.chaijs.com/api/bdd/#method_throw

And there are many other checks like this

test/Settings.ui.ts
```diff
@@ -100,20 +102,18 @@ describe('Settings', function () {
...
-            By.css('input[value="3P5Xx9MFs8VchRjfLeocGFxXkZGknm38oq1"]')
-          ),
-          this.wait
-        )
-        .click();
+      const item = await this.driver.wait(
+        until.elementLocated(By.css('[data-testid=exportMenuItem]')),
+        this.wait
+      );
+      await item.click();
```
faergeek, line 109: Does it make a difference if you `await .wait` and then `await .click`? As far as I understand there should be no difference https://www.selenium.dev/selenium/docs/api/javascript/module/selenium-webdriver/lib/webdriver_exports_WebElementPromise.html

faergeek (approved):

src/ui/components/pages/PermissionsSettings/components/list/List.tsx
```diff
@@ -23,7 +23,7 @@ export class List extends React.PureComponent<IProps> {
 
     return (
       <div className={styles.permissionList}>
-        {originsNames.map(name => (
+        {originsNames.sort().map(name => (
```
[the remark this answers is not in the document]
ehassed3: Because the sequence of sources is incorrectly displayed in this list due to the fact that sometimes the reference to the object is lost.

`const list = { 'b': 0, 'c': 1, 'a': 2 }
Object.keys(list) => ['b', 'c', 'a']`

But when the reference to the object is lost, an object with sorted fields by default is obtained instead:

`Object.keys({ 'a': 2, 'b': 0, 'c': 1 }) => ['a', 'b', 'c']`

src/ui/components/ui/pills/Pills.tsx
```diff
@@ -12,22 +12,20 @@ export class Pills extends React.PureComponent {
...
-                hidden={item.hidden}
-                selected={item.selected}
-                key={item.id}
-                {...props}
-              />
-            </CSSTransition>
-          ))}
-        </TransitionGroup>
-      </div>
+      <TransitionGroup className={myClassName}>
```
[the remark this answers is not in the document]
ehassed3: I'm just removing the redundant element.

src/ui/components/pages/PermissionsSettings/components/list/List.tsx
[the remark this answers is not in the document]
faergeek: Not sure what you mean when you say "reference to the object is lost". How is it lost?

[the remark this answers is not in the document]
ehassed3: When creating the `origins` object, the keys are written in the order that comes from the backend. Sometimes this object sorts its keys, most likely this is due to the loss of the original structure. I do not know why the reference to the object was lost, it only happens in chromedriver.

test/AccountCreation.ui.ts
smelnikov: When some transition comes, some pages/components are not reachable by `this.driver.findElement`, so we need to wait for them with `this.driver.wait`.
The original idea was that in some cases we need to check that an element appears successfully without breaking the test.
Now this check seems unnecessary.

test/utils/actions.ts
```diff
@@ -161,9 +167,57 @@ export const App = {
...
+    this.serviceWorkerTab = handles[handles.length - 1];
+
+    await this.driver.switchTo().window(extensionPanelHandle);
+    await this.driver.close();
+    await this.driver.switchTo().window(handles[0]);
+  },
+  restartServiceWorker: async function (this: mocha.Context) {
+    const currentHandle = await this.driver.getWindowHandle();
+
+    await this.driver.switchTo().newWindow('serviceworker');
```
smelnikov, line 207: This method accepts only `tab` or `window` value.

https://www.selenium.dev/documentation/webdriver/browser/windows/#create-new-window-or-new-tab-and-switch

test/utils/actions.ts
```diff
@@ -161,9 +167,57 @@ export const App = {
       this.wait
     );
   },
+  openServiceWorkerTab: async function (this: mocha.Context) {
+    await this.driver.switchTo().newWindow('extension');
```
smelnikov, line 171: This method accepts only `tab` or `window` value.

https://www.selenium.dev/documentation/webdriver/browser/windows/#create-new-window-or-new-tab-and-switch

smelnikov (approved):

PR #280, opened 2022-05-06
Dismount modals when locked

smelnikov:


ehassed3 (approved):

faergeek (approved):
Looks like `.hidden` class can be removed from styles also

PR #284, opened 2022-05-11
Eslint

smelnikov:


src/ui/components/ui/buttons/ApproveBtn.tsx
```diff
@@ -5,9 +5,23 @@ import { Button } from './Button';
...
+  disabled?: boolean;
+  loading?: boolean;
+  autoClickProtection?: boolean;
+  children: React.ReactNode;
+}
+
+export class ApproveBtn extends React.PureComponent<Props, State> {
   readonly props;
-  readonly state = {} as any;
+  readonly state = {} as State;
```
faergeek, line 24: `readonly state: State` is a bit better, because it doesn't cancel type checks

src/ui/components/ui/qRCode/QrCode.tsx
```diff
@@ -34,9 +46,24 @@ const QrCodeImage = ({ options, src, width, height, className, ...props }) => {
...
+  text?: string;
+  className?: string;
+  children?: React.ReactNode;
+};
+
 export class QRCode extends React.PureComponent {
-  readonly props;
-  readonly state = {} as any;
+  readonly props: QRCodeProps;
+  readonly state = {} as State;
```
faergeek, line 66: `readonly state: State`?

src/ui/components/pages/assets/tabs/tabAssets.tsx
```diff
@@ -13,8 +13,23 @@ import AutoSizer from 'react-virtualized-auto-sizer';
...
+}: {
+  data: {
+    assetEntries: Array<[string, { balance: string | number | BigNumber }]>;
+    assets: Record<string, AssetDetail>;
+    onInfoClick: (assetId: string) => void;
+    onSendClick: (assetId: string) => void;
+    onSwapClick: (assetId: string) => void;
+  };
+  index;
+  style;
```
faergeek, line 31: Forgot types for `index` and `style`)

src/ui/components/pages/SelectedAccountQr.tsx
```diff
@@ -70,7 +70,7 @@ class QRCodeSelectedAccountComponent extends React.PureComponent<Props> {
   }
 }
 
-const mapStateToProps = function (store: any) {
+const mapStateToProps = function (store) {
```
faergeek, line 73: `AppState`?

src/ui/components/pages/AccountInfo.tsx
```diff
@@ -412,7 +424,7 @@ class AccountInfoComponent extends React.Component<Props> {
   }
 }
 
-const mapStateToProps = function (store: any) {
+const mapStateToProps = function (store) {
```
faergeek, line 427: `AppState`?

src/ui/components/Root.tsx
```diff
@@ -173,7 +173,7 @@ class RootComponent extends React.Component {
   }
 }
 
-const mapStateToProps = function (store: any) {
+const mapStateToProps = function (store) {
```
faergeek, line 176: `AppState`?

faergeek (approved):
Good job 👍

scripts/WebpackCustomActionsPlugin.js
```diff
@@ -12,6 +12,7 @@ module.exports = class WebpackShellPlugin {
 
     compiler.hooks.beforeRun.tapPromise(
       'Custom actions on start',
+      /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
       (source, target, routesList) => {
```
ehassed3, line 16: It's better to remove unused vars

src/_sentryHelpersReplacement.js
```diff
@@ -131,4 +132,5 @@ export function wrap(fn, options, before) {
  * Injects the Report Dialog script
  * @hidden
  */
+// eslint-disable-next-line @typescript-eslint/no-empty-function
 export function injectReportDialog() {}
```
ehassed3, line 136: `{
    // do nothing.
}`

src/ui/components/RootAccounts.tsx
```diff
@@ -67,14 +67,14 @@ export function RootAccounts() {
...
-    [currentTab]
+    [currentTab, dispatch]
   );
 
   const onBack = React.useCallback(() => {
     const tab = backTabFromConf || backTabs[backTabs.length - 1] || PAGES.ROOT;
     dispatch(removeBackTab());
     dispatch(setTab(tab));
-  }, [backTabFromConf, backTabs]);
+  }, [backTabFromConf, backTabs, dispatch]);
```
ehassed3, line 77: You have a lot of fixes for the `exhaustive-deps` error, but I think we need to disable this rule. Adding additional parameters may be the cause of bugs and unnecessary deps.

ehassed3 (approved):

src/ui/components/RootAccounts.tsx
faergeek: Disabling this rule doesn't make sense. Often it's hard to track all the deps used in a hook.

What do you mean by "unnecessary deps"? `dispatch` is stable in this case. And it makes no difference if you add it or not, eslint just can't know for sure. I see no problem with that.

And regarding bugs. If some effect is called more often than needed, it's not that hard to fix. Are there any other kinds of bugs you have in mind?

I think this rule is very helpful.

scripts/getVersion.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
```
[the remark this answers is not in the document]
faergeek: In that case may be put it into `overrides`?
https://eslint.org/docs/user-guide/configuring/configuration-files#configuration-based-on-glob-patterns

src/ui/components/RootAccounts.tsx
ehassed3: I mean, the logic may not assume dependence on the added parameter, which may cause bugs. 

> eslint just can't know for sure

Therefore, this rule seems unnecessary to me. It is better to use it yourself than to write unnecessary deps for the sake of eslint.

ehassed3: Discussed in private messages, I think that the rule adds complexity to hooks usage but I agree that it generally reduces the number of bugs.

faergeek (approved):

ehassed3 (approved):

PR #285, opened 2022-05-11
Fix update favorite WAVES

ehassed3:


faergeek (approved):
The problem is in the reference to a single object, right? May be `WAVES` should also be copied to `defaults` in constructor instead of referencing it?

ehassed3:
> The problem is in the reference to a single object, right? May be `WAVES` should also be copied to `defaults` in constructor instead of referencing it?

No, the problem is overwriting the asset to the default value of `WAVES`.

ehassed3:
> The problem is in the reference to a single object, right? May be `WAVES` should also be copied to `defaults` in constructor instead of referencing it?

No, the problem is overwriting the asset to the default value of `WAVES`.

smelnikov (approved):

PR #286, opened 2022-05-13
Add flat local storage

ehassed3:


src/lib/localStore.ts
```diff
@@ -0,0 +1,188 @@
+const pump = require('pump');
+const log = require('loglevel');
+const debounceStream = require('debounce-stream');
+const asStream = require('obs-store/lib/asStream');
+const { extension } = require('lib/extension');
+const { createStreamSink } = require('lib/createStreamSink');
```
faergeek, line 6: Is there a problem with using `import` instead of `require`?

faergeek (approved):

smelnikov (approved):

PR #289, opened 2022-05-16
Fix button paddings

ehassed3:


src/ui/components/ui/buttons/Button.module.css
```diff
@@ -21,6 +21,10 @@
   border-radius: 0 4px 4px 0;
 }
 
+:global button.modal-close {
```
smelnikov, line 24: Maybe move it here?
https://github.com/wavesplatform/waves-keeper/blob/KEEP-354/fix-button-paddings/src/ui/styles/app.styl#L443

smelnikov (changes requested):

smelnikov (approved):

src/ui/components/ui/buttons/Button.module.css
smelnikov: Forgot the button element has padding

PR #294, opened 2022-05-19
Add other langs

ehassed3:


src/ui/components/Root.tsx
```diff
@@ -160,7 +160,7 @@ class RootComponent extends React.Component {
     const pageProps = { ...pageConf.props, setTab, onBack };
 
     return (
-      <div className="height">
+      <div className={`height ${this.props.currentLocale}`}>
```
smelnikov, line 163: Need to add the same to `ui/components/RootAccounts.tsx`

src/ui/styles/app.styl
```diff
@@ -99,6 +98,15 @@
...
   justify-content: center;
 
+:global .th *
+  font-family: Noto-Sans-Th, Roboto
+
+:global .ja *
+  font-family: Noto-Sans-Ja
+
+:global .zh *
+  font-family: Noto-Sans-Zh
```
smelnikov, line 108: It seems that `Noto-Sans-*`, fonts are taller.
Because of this, the dollar amount does not fit in the asset card.

smelnikov (approved):

PR #303, opened 2022-05-25
Add reverse migration flat state

ehassed3:


src/lib/localStore.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    await this._set({
+      migrationVersion: CURRENT_MIGRATION_VERSION,
+    });
+
+    if (version > CURRENT_MIGRATION_VERSION) {
+      for (let i = version; i > CURRENT_MIGRATION_VERSION; i--) {
+        const reverseMigrate = REVERSE_MIGRATIONS[i - 1];
+        if (reverseMigrate) {
+          await reverseMigrate.bind(this)();
```
faergeek, line 45: I guess it can be replaced with just `.call` instead of `.bind`ing and then calling it:

```typescript
await reverseMigrate.call(this);
```

src/lib/localStore.ts
```diff
@@ -0,0 +1,82 @@
+import * as extension from 'extensionizer';
+import log from 'loglevel';
+import { REVERSE_MIGRATIONS } from 'lib/reverseMigrations';
+
+const CURRENT_MIGRATION_VERSION = 0;
+
+export default class ExtensionStore {
+  public isSupported: boolean;
```
faergeek, line 8: May be remove this flag already? Keeper can't work without `storage.local`. It would be just useless without it :-)

faergeek (approved):
Actual reverse migration will be added in flat storage branch, right?

src/lib/localStore.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  async migrate() {
+    const { migrationVersion } = await this._get();
+    const version = (migrationVersion as number) || 0;
+
+    await this._set({
+      migrationVersion: CURRENT_MIGRATION_VERSION,
+    });
+
+    if (version > CURRENT_MIGRATION_VERSION) {
```
smelnikov, line 41: I think we need some fallback mechanism in case one of the `reverseMigrate` fails.

smelnikov (changes requested):

smelnikov (approved):

PR #316, opened 2022-06-07
Fix firefox dropdown button icon

ehassed3:


src/ui/components/ui/buttons/dropdownButton.styl
```diff
@@ -9,7 +9,7 @@
 
 .dropdownButton
   &::after
-    display: inline-block
+    display: block
```
smelnikov, line 12: By the way, the dropdown icon is shifted.

Looks like we just need to override `.button:not(.transparent):not(.custom)` in `Button.module.css` to use flex-box instead of block and remove `display` for pseudo-element.

smelnikov (changes requested):

smelnikov (approved):

PR #317, opened 2022-06-07
Fix moneylike burn transaction

ehassed3:


src/transactions/utils.ts
```diff
@@ -453,21 +453,33 @@ export const convertFromSa = {
...
       }
-      case TRANSACTION_TYPE.BURN:
+      case TRANSACTION_TYPE.BURN: {
+        const quantity = input.data.quantity || input.data.amount;
+
+        let assetId = input.data.assetId;
+
+        if (!assetId && quantity instanceof Money) {
+          assetId = quantity.asset.id;
+        }
```
smelnikov, line 463: If we have a MoneyLike `amount` and `assetId` are non-null, the `amount` asset id will be ignored.

Maybe simplify this to 
```javascript
const assetId =  quantity instanceof Money ? quantity.asset.id : input.data.assetId
```

smelnikov (changes requested):

PR #329, opened 2022-06-22
Support waves node 1.4.6

ehassed3:


package.json
```diff
@@ -53,6 +53,7 @@
     "debounce": "^1.2.1",
     "debounce-stream": "^2.0.0",
     "dnode": "^1.2.2",
+    "ethereumjs-util": "^7.1.5",
```
faergeek, line 56: Looks like this package has been deprecated according to repo https://github.com/ethereumjs/ethereumjs-util. As I understand it's moved here https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/util. Can we use the new one instead?

faergeek (approved):

package.json
ehassed3: No, we can't. `npm install @ethereumjs/util` throws an error and https://www.npmjs.com/package/@ethereumjs/util — not found

faergeek: Ok, no questions then

src/controllers/CalculateFeeController.js
```diff
@@ -44,20 +21,8 @@ const isAccountHasExtraFee = async (address, node) => {
...
 
-const getConfig = async function () {
-  const response = await fetch(DEFAULT_FEE_CONFIG_URL);
-  return await response.json();
-};
-
-const getCachingFeeConfig = promiseCache(
-  CONFIG_EXPIRATION_TIME,
-  DEFAULT_FEE_CONFIG
-)(getConfig);
```
smelnikov, line 55: Why don't we use remote config for fee values now?

smelnikov (approved):

src/controllers/CalculateFeeController.js
ehassed3: We decided at the meeting not to use remote config

PR #338, opened 2022-07-01
Add addresses not found message

ehassed3:


src/controllers/Statistics.js
```diff
@@ -98,7 +98,7 @@ export class StatisticsController {
             Accept: '*/*',
           },
           body: JSON.stringify({
-            api_key: statisticsApiKey,
+            api_key: config.statisticsApiKey || '',
```
smelnikov, line 101: May be don't `fetch` if `config.statisticsApiKey` is not set?

smelnikov (approved):

PR #339, opened 2022-07-01
Add invalid request data details

ehassed3:


smelnikov (approved):

src/lib/KeeperError.js
```diff
@@ -60,7 +60,8 @@ class KeeperError extends Error {
 export const ERRORS = Object.entries(ERRORS_DATA).reduce(
   (acc, [code, data]) => {
     const { msg, name } = data;
-    acc[name] = e => new KeeperError(msg, code, e);
+    acc[name] = (data, message) =>
```
smelnikov, line 63: May be swap these arguments?
The error is usually initialized like `new Error(message, metadata)`, so using `ERRORS_DATA.REQUEST_ERROR(object, message)` seems a bit confusing.

PR #355, opened 2022-07-11
Attachment as binary in

smelnikov:


faergeek (approved):

ehassed3 (approved):

src/transactions/utils.ts
```diff
@@ -1,12 +1,8 @@
...
 import { Money } from '@waves/data-entities';
 import { binary, serializePrimitives } from '@waves/marshall';
-import {
-  base58Encode,
-  blake2b,
-  concat,
-  stringToBytes,
-} from '@waves/ts-lib-crypto';
+import { base58Encode, blake2b, concat, TBinaryIn } from '@waves/ts-lib-crypto';
+import { _fromRawIn } from '@waves/ts-lib-crypto/conversions/param';
```
faergeek, line 5: I don't think it's a good idea to use these internal imports. May be just write our own checks directly?

PR #357, opened 2022-07-12
Add ui update optimization

ehassed3:


src/ui.tsx
```diff
@@ -59,7 +59,7 @@ async function startUi() {
     const stateChanges = await backgroundService.getState([
       'initialized',
       'locked',
-      'assets',
+      ...('currentNetwork' in changes ? ['assets'] : []),
```
faergeek, line 62: I constantly forget why we need to get these fields explicitly always. Guess we need an explanation in a code comment :-)

src/lib/localStore.ts
```diff
@@ -113,6 +116,22 @@ export default class ExtensionStore {
     );
   }
 
+  getState(keys?: string | string[]) {
```
smelnikov, line 119: May be use spread for argument? So this can be simplified to e.g.:
```typescript
getState(...keys?: string[]){
  if(keys.length === 0) return this._state;
  
  return keys.reduce(
      (acc, key) =>
        this._state[key] ? { ...acc, [key]: this._state[key] } : acc,
      {}
    );
}
```

ehassed3: The argument cannot be optional, and the typescript also complains because a string is not expected as an argument.

smelnikov: Ofс, spread arguments are always an array, and it should be `getState(...keys: string[])`.

PR #359, opened 2022-07-13
Fix amplitude events

ehassed3:


src/controllers/Statistics.js
```diff
@@ -80,13 +79,13 @@ export class StatisticsController {
     return this.sendEvents();
   }
 
-  sendEvents() {
+  sendEvents(time = 5000) {
```
smelnikov, line 82: May be rename `time` arg to `delay` or something else

smelnikov (approved):

PR #361, opened 2022-07-14
Fix tx fee options

faergeek:


src/fee/utils.ts
```diff
@@ -0,0 +1,124 @@
...
+import { DEFAULT_FEE_CONFIG } from '../constants';
+import { Message } from 'ui/components/transactions/BaseTransaction';
+import { AssetDetail } from 'ui/services/Background';
+import { AccountBalance, AssetBalance } from 'ui/reducers/updateState';
+import { SPONSORED_FEE_TX_TYPES } from './constants';
+
+export function convertFeeToAsset(fee: Money, asset: Asset) {
+  const minSponsoredFee = (asset: Asset) =>
+    asset.id === 'WAVES'
+      ? DEFAULT_FEE_CONFIG.calculate_fee_rules.default.fee
```
ehassed3, line 16: I think we need to use the config from https://raw.githubusercontent.com/Waves-Keeper/configs/master/fee.json. If we want to change the value of the field, then there will be a conflict between these two configs.

src/controllers/MessageController.js
```diff
@@ -698,14 +696,14 @@ export class MessageController extends EventEmitter {
           message.account
         );
 
-        const matcherFeeData = !hasFee
-          ? await this._getFee(message, result.data)
-          : {};
+        const matcherFee =
+          message.data.data?.matcherFee ||
```
ehassed3, line 700: `?.` will it work here?

faergeek: Why not? It didn't work earlier because of old babel. Now it works

src/fee/utils.ts
faergeek: Moved fee config to remote config controller

smelnikov (approved):

ehassed3 (approved):

PR #364, opened 2022-07-15
Add default assets

ehassed3:


faergeek (changes requested):
I don't think it's a good idea to fallback to testnet. I guess better not to dispatch an action at all if `assets[network] === undefined`, or dispatch with an empty object. But actually all of these solutions are not good. Better figure out the cause of the error first.

ehassed3:
An empty object also results in an error, because WAVES asset expected.

faergeek (approved):

PR #369, opened 2022-07-18
Add local storage backup

ehassed3:


src/lib/localStore.ts
```diff
@@ -26,10 +26,30 @@ const CONTROLLERS = [
   'VaultController',
 ];
 
+export async function backup() {
+  const { backup = {}, WalletController } = await extension.storage.local.get([
```
faergeek, line 30: No need for default value, `undefined` is ok to spread onto objects

faergeek (approved):

PR #374, opened 2022-07-20
Convert JS to TS

faergeek:


src/controllers/statistics.ts
```diff
@@ -154,14 +177,16 @@ export class StatisticsController {
      @param {number} [ms] in ms, by default is 1 hour
      @param {string} [storeKey] by default is `last{EventType}`
```
ehassed3, line 178: I think you need to remove JSDoc everywhere.

ehassed3 (approved):

smelnikov (approved):

PR #375, opened 2022-07-21
Update webpack

ehassed3:


src/ui.tsx
```diff
@@ -38,6 +38,7 @@ async function startUi() {
   const store = createUiStore();
 
   store.dispatch(setLangs(LANGS));
+  console.log();
```
faergeek, line 41: Should be removed I guess)

src/assets/amountInput.tsx
```diff
@@ -4,7 +4,7 @@ import * as React from 'react';
 import { useIMask } from 'react-imask';
 import { BalanceAssets } from 'ui/reducers/updateState';
 import { useAppSelector } from 'ui/store';
-import * as styles from './amountInput.module.css';
+import styles from './amountInput.module.css';
```
faergeek, line 7: May be enable `namedExport` for css-loader instead?

package.json
```diff
@@ -1,11 +1,25 @@
 {
   "private": true,
   "version": "2.9.0",
-  "engines": {
-    "node": "^10.24.1"
-  },
+  "browserslist": [
+    "maintained node versions",
```
faergeek, line 5: We don't want to compile for node at all

package.json
```diff
@@ -1,11 +1,25 @@
...
+    "supports es6-module and > 0.5%",
+    "not dead",
+    "not IE 11",
```
faergeek, line 8: I think everything below "not IE11" is unnecessary. May be just ignore safari explicitly. You can see which browser versions are targeted by adding `"debug": true` to `babel-preset-env` options.

webpack.common.js
```diff
@@ -0,0 +1,122 @@
+const path = require('path');
```
faergeek, line 1: I don't think it's a good idea to duplicate most parts of configs. They differ not so much and it makes it hard to understand what's going on. Plus there's just to much duplication between them now. I would put all config in just one file and tweak based on `mode` passed. @smelnikov what do you think?

src/lib/createStreamSink.js
```diff
@@ -25,10 +25,10 @@ function promiseToCallback(promise) {
   return function (cb) {
     promise.then(
       function (data) {
-        setImmediate(cb, null, data);
+        setTimeout(cb, 0, null, data);
```
faergeek, line 28: `setTimeout` is not the same as `setImmedate`. Actually I think none of them are needed here anyway. `Promise.then(...)` already creates a microtask. The same in `dnode-util`

package.json
```diff
@@ -99,37 +115,40 @@
...
+    "mini-css-extract-plugin": "^2.6.1",
+    "mini-svg-data-uri": "^1.4.4",
     "mocha": "^9.2.2",
     "prettier": "^2.4.1",
     "selenium-webdriver": "^4.1.2",
-    "source-map-loader": "^1.0.0",
-    "style-loader": "^0.22.1",
-    "stylelint": "^13.13.1",
+    "source-map-loader": "^4.0.0",
+    "ssestream": "^1.1.0",
```
faergeek, line 139: Was it meant only for implementing HMR?

faergeek (changes requested):

src/assets/amountInput.tsx
ehassed3: I don't think this is a good idea. It is better for styles to have one name.

faergeek: It's still one name, it will just export the same thing as before, so these changes would be unnecessary

ehassed3: But it's already fixed (:

faergeek: Okay, let's leave it changed then)

src/lib/dnode-util.js
```diff
@@ -62,7 +62,7 @@ export function promiseToCb(fn, context) {
       callback = noop;
     }
     fn.apply(context, args)
-      .then(result => setImmediate(callback, null, result))
-      .catch(error => setImmediate(callback, error));
+      .then(result => setTimeout(callback, 0, null, result))
```
faergeek, line 65: `setTimeout` is not needed here as well

faergeek (approved):

src/lib/createStreamSink.js
```diff
@@ -13,23 +13,13 @@ class AsyncWritableStream extends WritableStream {
...
-    throw new TypeError('Expected a promise');
-  }
-
-  return function (cb) {
-    promise.then(
-      function (data) {
-        setImmediate(cb, null, data);
+    this._asyncWriteFn(chunk, encoding).then(
+      () => {
+        callback();
```
smelnikov, line 18: Looks like you missed `null` and `data` parameters for `callback` function

smelnikov (approved):

src/global.d.ts
```diff
@@ -3,6 +3,8 @@ declare const __SENTRY_DSN__: string | undefined;
 declare const __SENTRY_ENVIRONMENT__: string | undefined;
 declare const __SENTRY_RELEASE__: string | undefined;
 
+declare module '*.module.css';
```
faergeek, line 6: Why is this needed?

PR #377, opened 2022-07-31
Keep 420 atqa framework launch

kenshirokavaguchi:


.gitignore
```diff
@@ -5,6 +5,11 @@
 
 /dist*
 /builds
+#atqa-framework
+/atqa-framework/output*
+/atqa-framework/node_modules*
```
faergeek, line 10: node_modules are already ignored below

.gitignore
```diff
@@ -5,6 +5,11 @@
 
 /dist*
 /builds
+#atqa-framework
+/atqa-framework/output*
```
faergeek, line 9: Why do you need a `*` here? Do you ignore just `output` folder or also some `outputblabla`?

.gitignore
```diff
@@ -5,6 +5,11 @@
...
+/atqa-framework/node_modules*
+/atqa-framework/.env
+/atqa-framework/yarn-error.log
```
faergeek, line 12: All `*.log` files are already ignored below

atqa-framework/jsconfig.json
```diff
@@ -0,0 +1,5 @@
+{
```
faergeek, line 1: Why `jsconfig.json` is needed? Isn't `tsconfig.json` enough?

src/ui/components/pages/exportAccounts/passwordModal.tsx
```diff
@@ -26,6 +26,7 @@ export function ExportPasswordModal({
 
   const [password, setPassword] = React.useState('');
   const [passwordError, setPasswordError] = React.useState(false);
+  const [encrypted, setEncrypted] = React.useState(true);
```
faergeek, line 29: What's this?

atqa-framework/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "ts-node": {
+    "files": true
+  },
+  "compilerOptions": {
+    "target": "es2018",
+    "lib": ["es2016", "es2017", "es2018", "es2019", "es2020", "dom"],
```
faergeek, line 7: AFAIK you don't need to specify every version, only the latest one

atqa-framework/codecept.conf-FIREFOX.js
```diff
@@ -0,0 +1,58 @@
+require('ts-node/register');
+// const fs = require('fs-extra')
+//const FirefoxProfile = require('firefox-profile');
+// const extPath = fs.readFileSync('/Users/a1/waves/atqa-waves/extension/keeper-wallet-2.8.1-chrome.crx', {encoding: "base64"});
```
faergeek, line 4: May be remove commented out lines?

atqa-framework/codecept.conf-FIREFOX.js
```diff
@@ -0,0 +1,58 @@
...
+      //    version: true,
+      //    seleniumArgs: ['-host', '127.0.0.1','-port', '5555']
+      //},
+      //capabilities: {
+      //    browserName: 'firefox',
+      //    port: 5555,
+      //    'xpinstall.signatures.required': false,
+      //},
+      firefoxProfile: {
+        extensions: ['/Users/a1/waves/atqa-waves/extension/firefox'],
```
faergeek, line 46: I guess it's you local path. How would it work on other machines?

atqa-framework/browsers.json
```diff
@@ -0,0 +1,49 @@
+{
+  "chrome": {
+    "default": "102.0",
+    "versions": {
+      "102.0": {
+        "image": "selenoid/vnc_chrome_wv:102.0",
+        "port": "4444",
+        "path": "/",
+        "screenResolution": "1920x1080x24",
+        "volumes": ["/Users/a1/waves/waves-keeper/dist:/ext/tmp"]
```
faergeek, line 10: Local paths

atqa-framework/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM selenoid/vnc_opera:88.0
```
faergeek, line 1: For what is this `Dockerfile` used?

faergeek (changes requested):

faergeek:
What about adding a workflow to run this in GitHub Actions?

atqa-framework/tests/pages/AccountPage.ts
```diff
@@ -0,0 +1,20 @@
+import { BasePage } from './BasePage';
+
+export class AccountPage extends BasePage {
```
smelnikov, line 3: As far as I know, the idea of page object is to provide an interface for interacting with elements, e.g:
```typescript
// in test
accountPage = AccountPage(I)
accountPage.clickGetStartedButton()
accountPage.setBothPasswords('some_password_value")
// etc..

```
But this complicates some aspects of development (you need to support api of page object)

I think in this case we can simplify the interface and store selectors as class attributes, for example:
```typescript
class AccountPage extends BasePage{
    getStartedBtn = locate('button').withText('Get Started')
    passwordCreateInput = { xpath: '(//input)[1]' }
    passwordConfirmInput = { id: 'second' }
    continueBtn = locate('button').withText('Continue')
    // etc..
}

File: atqa-framework/tests/integration/scenario.Test.ts [hunk 12]
```diff
@@ -0,0 +1,68 @@
...
+});
+
+Scenario('Init scenario', async () => {
+  //I.wait(clockUnit.MINUTES * 30);
+  I.waitForEnabled(
+    accountPage.SELECTORS.GET_STARTED_BUTTON,
+    clockUnit.SECONDS * 30
+  );
+  I.doubleClick(accountPage.SELECTORS.GET_STARTED_BUTTON);
+  I.wait(clockUnit.SECONDS * 10);
```
smelnikov, line 26: Using `wait` is bad pattern.

atqa-framework/tests/integration/scenario.Test.ts
```diff
@@ -0,0 +1,68 @@
...
+
+Background(async () => {
+  return extensionInitHandler.extensionInit();
+});
+
+Scenario('Init scenario', async () => {
+  //I.wait(clockUnit.MINUTES * 30);
+  I.waitForEnabled(
+    accountPage.SELECTORS.GET_STARTED_BUTTON,
+    clockUnit.SECONDS * 30
```
smelnikov, line 23: `30 * clockUnit.SECONDS` is easier to understand

atqa-framework/tests/integration/scenario.Test.ts
```diff
@@ -0,0 +1,68 @@
+import { ClockUnit } from '../../utils/clockUnit';
+import { BasePage } from '../pages/BasePage';
+import { ExtensionInitHandler } from '../../utils/ExtensionInitHandler';
+import { AccountPage } from '../pages/AccountPage';
+
+const basePage = new BasePage();
```
smelnikov, line 6: Unused variable

atqa-framework/tests/pages/ExtensionInitPage.ts
```diff
@@ -0,0 +1,23 @@
+import { BasePage } from './BasePage';
+
+export class ExtensionInitPage extends BasePage {
+  public SELECTORS = {
```
smelnikov, line 4: There is no need to explicitly specify the `public` data modifier for class attributes/methods, it is used by default

atqa-framework/tests/pages/BasePage.ts
```diff
@@ -0,0 +1,117 @@
...
+      DETAILS_BUTTON: { xpath: '//span[contains(text(),"Details")]' },
+      DEV_MODE_TOGGLE: { css: '#developer-mode' },
+      EXTENSION_MENU_BUTTON: {
+        xpath: '(//button[@aria-label= "Extensions menu"])[2]',
+      },
+    },
+    FIREFOX: {
+      MANIFEST_EMAIL_TEXT: 'support@wavesplatform.com',
+    },
+  };
```
smelnikov, line 46: Nested selectors are harder to understand.
Also, mixing them with constants is a bad idea.

atqa-framework/tests/integration/scenario.Test.ts
kenshirokavaguchi: This scenario for debugging.

atqa-framework/tests/pages/BasePage.ts
kenshirokavaguchi: The structure isn't finished yet, maybe it is a reason for some confusion. I'm still working on it and will be more clear in the feature. But just in case, nested selectors is one the idea of POM it helps describe a page with correct and flexible structure instead declare it as a list. Selectors list do nothing, the flat structure is nicy when u are testing 1-page web application (but u don't need any pom in this way). But if your app has a complicated logic, lot of the pages and elements on it so this should be describe as structure for efficient work with that.

atqa-framework/tests/pages/ExtensionInitPage.ts
kenshirokavaguchi: fixed

.gitignore
faergeek: `/atqa-framework/output` is enough for that

atqa-framework/jsconfig.json
faergeek: `jsconfig.json` is needed only when there's no `tsconfig.json`

atqa-framework/tsconfig.json
faergeek: `es2020` includes `es2019`, `es2019` includes `es2018`, they're based on top of each other. You can just specify `es2022` here and in target

atqa-framework/tests/pages/AccountPage.ts
smelnikov: If we made some structural changes to a complex component/page, you need to update the nested structure for the page object, this can be tricky.

My idea is to have a page object class that contains:
* unambiguous element names
* a locator to find it on that page

atqa-framework/tests/integration/scenario.Test.ts
smelnikov: Okay, but don't forget to delete it afterwards.

atqa-framework/tests/pages/BasePage.ts
smelnikov: As far as I know the idea of POM is to define a page object with locators and methods for interacting with it.
POM doesn't imply nesting of objects. Many pages - many POM objects.

You can see some examples here
https://www.selenium.dev/documentation/test_practices/encouraged/page_object_models/

.gitignore
kenshirokavaguchi: sorry for confusing, we have to keep an output folder and ignore only content inside cause this folder defined in the configs for the report data

atqa-framework/tests/pages/AccountPage.ts
kenshirokavaguchi: the nested structure needs to provide more clear situation around exact pages cause locator which has been named just **locatorPasswordButton** - means nothing and hard to future maintaining and it also touches for list

but locator object 

`AUTH_FORM: {
      PASSWORD_INPUT: '#input_1',
      CONFIRM_PASSWORD_INPUT: '#input_2'
}
`
means a lot, a lot more than just **passwordInput**, it also provide an ability to group locators objects within one page 

the pages contains a lot of logic and locators should be grouped and structured thus for more easier interact with each page.

upper case makes such objects unique and provide a 100% understanding what is exactly does such object which could be declared in any place of framework.

it doesn't make developing harder at all but on the contrary, it allows you to simplify page navigation and code maintaining

atqa-framework/tests/pages/BasePage.ts
kenshirokavaguchi: check my comments above

.gitignore
faergeek: So if you remove the folder, it can't create it? Sounds strange

PR #392, opened 2022-08-10
Fix Infura endpoint for Sign Art

smelnikov:


src/nfts/index.ts
```diff
@@ -23,20 +24,28 @@ export enum NftVendor {
...
 export interface BaseInfo {
   id: string;
   vendor: NftVendor;
 }
 
 export class BaseNft<TypedInfo extends BaseInfo> {
   protected asset: AssetDetail;
-  protected info: TypedInfo | null;
+  protected info: TypedInfo | null | undefined;
+  protected config: NftConfig | null | undefined;
```
faergeek, line 35: If prop type matches what it's assigned inside constructor, you can omit types here.

src/nfts/ducks/index.ts
```diff
@@ -10,14 +10,23 @@ import {
...
-  constructor(asset: AssetDetail, info: DuckInfo) {
-    super(asset, info);
+  constructor({
+    asset,
+    info,
+    config,
+  }: {
+    asset: AssetDetail;
+    info?: DuckInfo | null;
+    config?: NftConfig | null;
```
faergeek, line 27: Doesn't seem to be ok that we can have no config for nfts. We have a default value anyway, right?

faergeek (approved):

src/nfts/ducks/index.ts
smelnikov: Of course, config is always set with the default value in the redux state.
Just a mistake.

PR #395, opened 2022-08-12
Improve performance

faergeek:


src/controllers/currentAccount.ts
```diff
@@ -71,20 +86,114 @@ export class CurrentAccountController {
...
+  async #fetchAliases(address: string) {
+    const url = new URL(`alias/by-address/${address}`, this.getNode());
+
+    const response = await fetch(url, {
+      headers: {
+        accept: 'application/json',
+      },
+    });
+
+    const json = (await response.json()) as unknown[];
```
smelnikov, line 168: This response is an array of strings.

src/storage/migrations.ts
```diff
@@ -121,7 +121,53 @@ const removeCurrentNetworkAccounts: Migration = {
   },
 };
 
+const flattenBalances: Migration = {
```
smelnikov, line 124: Do we really need to transfer the balance data, because they will still be fetched by the background task?!

smelnikov (approved):

src/storage/migrations.ts
faergeek: Why not?

PR #398, opened 2022-08-13
Close accounts tab with the finish button in firefox

smelnikov:
The firefox security policy does not allow a `window.close()` if it is not opened by a script

### DIFF
--- a/src/lib/tabsManager.ts
+++ b/src/lib/tabsManager.ts
@@ -40,4 +40,13 @@ export class TabsManager {
       })
     );
   }
+
+  async closeCurrentTab() {
+    extension.tabs.query(
+      { active: true, lastFocusedWindow: true },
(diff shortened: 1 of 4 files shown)


src/lib/tabsManager.ts
```diff
@@ -40,4 +40,13 @@ export class TabsManager {
       })
     );
   }
+
+  async closeCurrentTab() {
+    extension.tabs.query(
```
faergeek, line 45: Let's use promises instead of callbacks

faergeek (changes requested):

faergeek (approved):

src/lib/tabsManager.ts
smelnikov: Manifest v2 does not support promises.

PR #421, opened 2022-10-12
Use one config instead of three

faergeek:


smelnikov (approved):
Maybe we should combine the main config and legacy? How time-consuming is that?

faergeek:
I created a separate task to get rid of legacy config. It's better to figure out what parts of it are used and where, and then combine only parts that we actually need

KirillZamotaev (approved):

PR #568, opened 2022-12-20
KEEP-769: fix endless loading balances

domnikov-timofei:


src/controllers/currentAccount.ts
```diff
@@ -325,4 +325,47 @@ export class CurrentAccountController {
...
+    const currentNetwork = this.getNetwork();
+    const url = new URL('addresses/balance', this.getNode());
+    const addresses = this.getAccounts().map(account => account.address);
+
+    while (addresses.length > 0) {
+      const splicedAddresses = addresses.splice(0, 1000);
+      const response = await fetch(url, {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
```
faergeek, line 339: It's better to add `accept` header as well, so fields with 64-bit integers will be returned as strings instead of numbers:

```js
accept: 'application/json; large-significand-format=string',
```

There are examples of this right in this file.
https://docs.waves.tech/en/waves-node/node-api/headers

src/controllers/currentAccount.ts
```diff
@@ -325,4 +325,47 @@ export class CurrentAccountController {
...
+          addresses: splicedAddresses,
+        }),
+      });
+
+      const regularBalances = (await response.json()) as Array<{
+        id: string;
+        balance: string;
+      }>;
+
+      const balances = regularBalances.reduce<Record<string, BalancesItem>>(
```
faergeek, line 351: May be better use `map` + `Object.fromEntries` instead? It's easier to understand and I guess it's also more efficient than spreading object on every iteration :-)

src/controllers/currentAccount.ts
```diff
@@ -325,4 +325,47 @@ export class CurrentAccountController {
...
+        (acc, regularBalance) => {
+          const balanceKey = `balance_${regularBalance.id}`;
+          const existingBalance = this.store.getState()[balanceKey];
```
faergeek, line 354: May be separate `this.store.getState()` and getting a key in it? It would be more readable. And this way it's trivial to call `this.store.getState()` just once outside of the loop, instead of calling it on every iteration.

src/controllers/currentAccount.ts
```diff
@@ -325,4 +325,47 @@ export class CurrentAccountController {
...
+
+          const balance = existingBalance
```
faergeek, line 356: `undefined` spreads onto an object pretty well, so there's no need for this ternary

src/ui/components/pages/otherAccounts.tsx
```diff
@@ -44,10 +44,25 @@ export function OtherAccountsPage() {
...
+      typeof balance?.regular !== 'undefined'
+        ? new Money(balance.regular, wavesAsset)
+        : undefined,
     ])
   );
 
+  useEffect(() => {
+    background.updateOtherAccountsBalances();
+
+    const intervalId = setInterval(
```
faergeek, line 56: I thought we discussed that we'd better start next request after previous one is done. Am I wrong? If not, then `setInterval` should be replaced with `setTimeout`. And `setTimeout` should be called only after the update is done.

faergeek (changes requested):

smelnikov (approved):

faergeek (approved):

.size-limit.json
```diff
@@ -16,7 +16,7 @@
       "dist/build/popup.js",
       "dist/build/vendors*.js"
     ],
-    "limit": "529 kB"
+    "limit": "561 kB"
```
faergeek, line 19: It didn't really increase that much, right?

PR #586, opened 2022-12-28
KEEP-538: get list of supported assets from Aggregator API

domnikov-timofei:


src/ui/components/pages/swap/form.tsx
```diff
@@ -410,59 +409,58 @@ export function SwapForm({
     }
   }, [swapVendorTouched, profitVendor]);
 
+  const swappableAssetEntries = usePopupSelector(
```
smelnikov, line 412: Maybe rename it to `swappableAssetIdsByVendor`?
The current name is a bit confusing.

smelnikov (approved):

src/ui/components/pages/swap/form.tsx
```diff
@@ -410,59 +409,58 @@ export function SwapForm({
...
         return {
           ...asset,
           disabled: !isAvailable,
           disabledTooltip: isAvailable
             ? undefined
             : t('swap.notSwappablePairTooltip', {
-                assetTicker1: fromAsset.ticker,
-                assetTicker2: asset.ticker,
+                assetDisplayName1: fromAsset.ticker || fromAsset.displayName,
+                assetDisplayName2: asset.ticker || asset.displayName,
```
faergeek, line 459: `displayName` is already `ticker || name`, no need for `asset.ticker || ` here

src/ui/components/pages/swap/form.tsx
faergeek: I would also say that redux state and storage key should be renamed as well

src/ui/components/pages/assets/assetItem.tsx
```diff
@@ -174,7 +175,7 @@ export function AssetItem({
           </Tooltip>
 
           {currentNetwork === 'mainnet' &&
-            isSwappableAsset(currentNetwork, assetId) && (
+            Object.values(swappableAssets).flat().includes(assetId) && (
```
faergeek, line 178: May be pass `isSwappable` from the outside? Otherwise for every asset item there's a loop, which first flattens arrays and then searches for asset id. Also when passing from outside (asset list) it would make sense to convert it to set first, to utilize `Set#has` method, which I beleive is `O(1)`, not `O(n)`

.size-limit.json
```diff
@@ -16,7 +16,7 @@
       "dist/build/popup.js",
       "dist/build/vendors*.js"
     ],
-    "limit": "529 kB"
+    "limit": "559 kB"
```
faergeek, line 19: Seems like size limit for ui is decreased actually

faergeek (approved):

faergeek (approved):

PR #638, opened 2023-01-20
KEEP-1015: SignWrapper refactoring

domnikov-timofei:


src/ui/components/pages/importEmail/signContext.tsx
```diff
@@ -0,0 +1,161 @@
+import clsx from 'clsx';
```
faergeek, line 1: I think this file should be moved into a more generic place, like `messages` folder may be or `_core`. `signWrapper` was here because we initially implemented `email` and `ledger` on top of each other and forgot to move it out of here :-)

src/ui/components/pages/importEmail/signContext.tsx
```diff
@@ -0,0 +1,161 @@
...
+  useRef,
+  useState,
+} from 'react';
+import { useTranslation } from 'react-i18next';
+import { Button } from 'ui/components/ui/buttons/Button';
+import { Modal } from 'ui/components/ui/modal/Modal';
+import Background from 'ui/services/Background';
+
+import { Login } from './login';
+import * as styles from './signWrapper.module.css';
```
faergeek, line 20: Better rename css file according to this file's name

src/ui/components/pages/importEmail/signContext.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+export const SignContext = createContext<{
+  openConfirmModal: () => Promise<unknown>;
+  closeConfirmModal: () => void;
+}>({
+  openConfirmModal: async () => undefined,
+  closeConfirmModal: () => undefined,
```
faergeek, line 27: I think that in general instead of passing noops as initial value it's better to make it null, but add an invariant to `useSign` hook. So in case there's no context (we're trying to call it upper in the tree) we'll get a runtime error instead of silently doing nothing.

src/ui/components/pages/importEmail/signContext.tsx
```diff
@@ -0,0 +1,161 @@
...
+          onConfirm(params);
+          break;
+      }
+
+      setPending(false);
+    },
+    [account, onConfirm, openConfirmModal, closeConfirmModal]
+  );
+
+  return [sign, pending];
```
faergeek, line 160: I think it makes sense to use an object instead of a tuple here, so we decide about names right here. I see no reason to allow giving these values arbitrary names at a call site. It's not a generic thing, like e.g. `useState`

src/ui/components/pages/importEmail/signContext.tsx
```diff
@@ -0,0 +1,161 @@
...
+  openConfirmModal: async () => undefined,
+  closeConfirmModal: () => undefined,
+});
+
+export const SignProvider = ({ children }: { children: ReactNode }) => {
+  const { t } = useTranslation();
+
+  const account = usePopupSelector(state => state.selectedAccount);
+
+  const [modalOpen, setModalOpen] = useState(false);
```
faergeek, line 35: May be rename to `isModalOpen`?

src/ui/components/pages/importEmail/signContext.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+  const modalPromiseRef = useRef<{
+    resolve: null | ((value: unknown) => void);
+  }>({ resolve: null });
```
faergeek, line 39: Seems like making the whole object nullable makes more sense. Or may be not use an object, just keep resolve fn itself and name it `modalResolveRef`

src/ui/components/pages/importEmail/signContext.tsx
```diff
@@ -0,0 +1,161 @@
...
+export function useSign<OnConfirmParams>(
+  onConfirm: (params: OnConfirmParams) => void | Promise<void>
+): [(params: OnConfirmParams) => Promise<void>, boolean] {
+  const account = usePopupSelector(state => state.selectedAccount);
+
+  const [pending, setPending] = useState(false);
+
+  const { openConfirmModal, closeConfirmModal } = useContext(SignContext);
+
+  const sign = useCallback(
```
faergeek, line 121: Can this function be passed from provider instead, so all this logic for wallet types will be closer to each other? And if it's possible to get rid of ref keeping resolve function in provider that would be really great, because it's sometimes hard to follow the code with these kinds of things involved

faergeek (changes requested):
At the end, it seems like it would be better if `useSign` hook would just get value from context, throw if it's `null` and if not, pass the value outside. And all the logic will be implemented inside provider

src/ui/components/pages/swap/swap.tsx
```diff
@@ -97,223 +97,219 @@ export function Swap() {
...
-          onSwap={onPrepare}
-        />
-      )}
-    </SignWrapper>
+      }
+    },
+    [assets, selectedAccount, t, wavesFeeCoins]
+  );
+
+  const { sign, pending: signPending } = useSign(onConfirm);
```
faergeek, line 286: May be name it `signPending` (or better `isSignPending`) in `useSign` instead of renaming on the call site? It was the whole point of changing from array to object as a return value, to have better name already chosen for us :-)

faergeek (approved):

faergeek (approved):

PR #650, opened 2023-01-26
load usd prices from data service

domnikov-timofei:


src/ui/components/ui/UsdAmount/index.tsx
```diff
@@ -8,18 +12,22 @@ interface Props {
...
     return null;
   }
 
-  return !usdPrices[id] || usdPrices[id] === '1' ? null : (
-    <p className={className}>{`≈ $${new BigNumber(usdPrices[id])
-      .mul(tokens)
-      .toFixed(2)}`}</p>
+  return (
+    <p className={className}>
+      {`≈ $${new BigNumber(usdPrices[id]).mul(tokens).toFixed(2)}`}
```
faergeek, line 30: Seems like using jsx for formatting would be easier to understand here

src/ui/components/ui/UsdAmount/index.tsx
```diff
@@ -8,18 +12,22 @@ interface Props {
...
   const isMainnet = currentNetwork === 'mainnet';
 
-  if (!usdPrices || !isMainnet) {
+  const usdPrices = useUsdPrices(useMemo(() => [id], [id]));
+
+  if (typeof usdPrices[id] !== 'string') {
+    return <Loader />;
+  }
+
+  if (!isMainnet || !usdPrices[id] || usdPrices[id] === '1') {
```
faergeek, line 24: You already check for `usdPrices[id]` above. Checking that it's a string is enough. If it's empty, then it's a backend problem. And we don't need a special case for '1' value, we decided to show usd prices for all assets always.

src/controllers/assetInfo.ts
```diff
@@ -4,7 +4,10 @@ import { NetworkName } from 'networks/types';
 import ObservableStore from 'obs-store';
 import Browser from 'webextension-polyfill';
 
-import { defaultAssetTickers, stablecoinAssetIds } from '../assets/constants';
+import {
+  defaultAssetTickers,
+  // stablecoinAssetIds
```
faergeek, line 9: Commented out import, should be removed

src/controllers/assetInfo.ts
```diff
@@ -392,49 +381,36 @@ export class AssetInfoController {
...
-              ticker.priceAssetID === MARKETDATA_USD_ASSET_ID
-            ) {
-              acc[ticker.amountAssetID] = ticker['24h_close'];
-            }
-
-            return acc;
-          },
-          {}
-        );
+    if (!assetIds.length || network !== NetworkName.Mainnet) {
```
faergeek, line 387: I know that `!assetIds.length` and `assetIds.length === 0` check the same thing, but with the latter the intent is much clearer, don't you agree? :-)

src/_core/polling.ts
```diff
@@ -1,6 +1,6 @@
 export function startPolling(
   ms: number,
-  fn: (abortSignal: AbortSignal) => Promise<void>
+  fn: (abortSignal: AbortSignal) => Promise<unknown>
```
faergeek, line 3: Is this change really needed? We don't expect function to return anything, because it's a side-effect function, so I think it's better to stick to `void` here

src/_core/usdPrices.tsx
```diff
@@ -0,0 +1,71 @@
...
+} from 'react';
+import invariant from 'tiny-invariant';
+
+import { usePopupSelector } from '../popup/store/react';
+import Background from '../ui/services/Background';
+import { startPolling } from './polling';
+import { useDebouncedValue } from './useDebouncedValue';
+
+const UsdPricesContext = createContext<
+  null | ((assetIds: string[]) => undefined | (() => void))
```
faergeek, line 17: Values like `null` & `undefined` seems to be much more expected at the end of a union, not at the beginning)

src/_core/usdPrices.tsx
```diff
@@ -0,0 +1,71 @@
...
+>(null);
+
+export function UsdPricesProvider({ children }: { children: ReactNode }) {
+  const [observedAssetIds, setObservedAssetIds] = useState<string[][]>([]);
+  const observedAssetIdsDebounced = useDebouncedValue(observedAssetIds, 100);
+
+  useEffect(() => {
+    if (!observedAssetIdsDebounced.length) {
```
faergeek, line 25: Same thing, `!<something>.length` vs `<something>length === 0`

src/_core/usdPrices.tsx
```diff
@@ -0,0 +1,71 @@
...
+      return;
+    }
+
+    return startPolling(5000, () =>
+      Background.updateUsdPricesByAssetIds([
+        ...new Set(observedAssetIdsDebounced.flat()),
```
faergeek, line 31: I doubt any other items would ever be added to this array, so may be use `Array.from` instead of spreading onto an empty array?

src/_core/usdPrices.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const observe = useCallback((assetIds: string[]) => {
+    if (!assetIds.length) {
+      return;
+    }
+
+    setObservedAssetIds(ids => [...ids, assetIds]);
+
+    return () => {
+      setObservedAssetIds(ids => {
+        ids.splice(ids.indexOf(assetIds), 1);
```
faergeek, line 45: Mutating state doesn't seem to be a good idea. May be use `.filter` instead?

src/_core/usdPrices.tsx
```diff
@@ -0,0 +1,71 @@
...
+  invariant(observeUsdPricesByAssetIds);
+
+  const usdPrices = usePopupSelector(state => state.usdPrices);
+
+  useEffect(
+    () => observeUsdPricesByAssetIds(assetIds),
+    [observeUsdPricesByAssetIds, assetIds]
+  );
+
+  return usdPrices;
```
faergeek, line 70: May be filter returned object to only have `assetIds` keys? Otherwise you can write `useUsdPrices([])` and get all the values, except they won't be kept fresh. Which in turn can be a debugging nightmare.

And probably wrap it into a `useMemo` then.

faergeek (changes requested):

smelnikov (changes requested):

faergeek (approved):

src/ui/components/ui/UsdAmount/index.tsx
```diff
@@ -8,18 +12,22 @@ interface Props {
...
   const isMainnet = currentNetwork === 'mainnet';
 
-  if (!usdPrices || !isMainnet) {
+  const usdPrices = useUsdPrices(useMemo(() => [id], [id]));
+
+  if (typeof usdPrices[id] !== 'string') {
+    return <Loader />;
+  }
+
+  if (!isMainnet || usdPrices[id] === '1') {
```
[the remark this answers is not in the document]
domnikov-timofei: it was already here, but yes, it seems unnecessary

faergeek (approved):

smelnikov (approved):

PR #668, opened 2023-01-31
handle data-entries limit

domnikov-timofei:


src/_core/chunk.ts
```diff
@@ -0,0 +1,13 @@
+export function chunk<T>(array: T[], count: number) {
```
smelnikov, line 1: Is there any reason why we can't use `slice` as in this case?
https://github.com/Keeper-Wallet/Keeper-Wallet-Extension/blob/KEEP-1063/handle-data-entries-limit/src/controllers/assetInfo.ts#L352:L367

src/nfts/utils.ts
```diff
@@ -4,12 +4,11 @@ export function capitalize(str: string | undefined) {
...
-export function reduceDataEntries(entries: DataTransactionEntry[]) {
-  return entries.reduce<Record<string, DataTransactionEntry['value']>>(
-    (data, item) => {
-      data[item.key] = item.value;
-      return data;
-    },
-    {}
-  );
+export function reduceDataEntries<
+  T extends { key: DataTransactionEntry['key']; value: unknown }
```
smelnikov, line 8: Actually, the `entries` type is `DataTransactionEntry[]`.
Why are we omitting `type`?

src/_core/chunk.ts
domnikov-timofei: maybe something like this?

```javascript
function* chunk(array, size) {
  for (const i = 0; i < array.length; i += size) {
    yield array.slice(i, i + size);
  }
}
```

seems to be easier to read)

src/nfts/utils.ts
domnikov-timofei: `DataTransactionEntry` is a union type that combines `DataTransactionEntryInteger`, `DataTransactionEntryString` and so on. if we pass data of type such as `DataTransactionEntryInteger` to this function, then at the output we will get a type inferred from `DataTransactionEntry`, but we expected it to be processed as `DataTransactionEntryInteger`

src/_core/fetchInBatches.ts
```diff
@@ -0,0 +1,39 @@
+import invariant from 'tiny-invariant';
+
+import { chunk } from './chunk';
+
+export function fetchInBatches<T>(
+  allItems: unknown[],
```
faergeek, line 6: Why not use a generic for this as well instead of `unknown`?

src/_core/fetchInBatches.ts
```diff
@@ -0,0 +1,39 @@
...
+  chunkSize: number,
+  fetchFn: (items: unknown[]) => ReturnType<typeof fetch>
+): Promise<T[]> {
+  const chunks = chunk(allItems, chunkSize);
+
+  // eslint-disable-next-line no-async-promise-executor
+  return new Promise<T[]>(async (resolve, reject) => {
```
faergeek, line 13: Why do you need a promise constructor here at all? Why not just turn this whole `fetchInBatches` function into an async function instead?

This way you won't need to wrap loop contents in a try/catch and `reject` and `break` in `catch`. Let it just throw instead :-)

src/_core/fetchInBatches.ts
```diff
@@ -0,0 +1,39 @@
+import invariant from 'tiny-invariant';
+
+import { chunk } from './chunk';
+
+export function fetchInBatches<T>(
+  allItems: unknown[],
+  chunkSize: number,
+  fetchFn: (items: unknown[]) => ReturnType<typeof fetch>
+): Promise<T[]> {
```
faergeek, line 9: `fetchFn` should return `Promise<T[]>` instead. This function doesn't need to know anything about `fetch`

src/nfts/utils.ts
faergeek: Just `T extends DataTransactionEntry` would work as well

src/nfts/vendors/signArt.ts
```diff
@@ -124,7 +126,8 @@ export class SignArtNftVendor implements NftVendor<SignArtNftInfo> {
           const artDisplayCid = artworksEntries[entriesPerAsset * index + 2];
           const userName = userNameEntries[index];
           const [, artworkId, creator] =
-            artName.key.match(/art_name_(\w+)_(\w+)/i);
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            artName.key.match(/art_name_(\w+)_(\w+)/i)!;
```
faergeek, line 130: Better use `invariant` instead of type assertion

faergeek (changes requested):
I think it would be better to implement a function like `fetchDataEntries`, which will:

1. Use `fetchInBatches` internally. Node api limits for data entries doesn't change between different requests anyway, so we don't need to pass it every time.
2. Do what `reduceDataEntries` does, meaning it will always return a record back instead of an array.

What do you think?

faergeek (approved):

smelnikov (approved):

PR #669, opened 2023-02-01
Add Update.ts test

denis-y-krasilnikov:


.github/workflows/update-tests.yml
```diff
@@ -0,0 +1,51 @@
...
+on:
+  push:
+    branches-ignore:
+      - master
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: styfle/cancel-workflow-action@0.11.0
+      - run: sudo chown -R $USER:$USER $GITHUB_WORKSPACE
```
faergeek, line 11: Seems like this `chown` is not actually needed. It was in `tests.yml`, but since we run `docker compose` directly it's not needed anymore, so I removed it (https://github.com/Keeper-Wallet/Keeper-Wallet-Extension/pull/667/commits/cd4e8e13976f213620f1e743d0ac19f1fc422483)

.github/workflows/update-tests.yml
```diff
@@ -0,0 +1,51 @@
...
+      # Build current branch
+      - name: Get Current branch source
+        uses: actions/checkout@v3
+        with:
+          clean: false
```
faergeek, line 16: `clean` is unnecessary on first checkout

.github/workflows/update-tests.yml
```diff
@@ -0,0 +1,51 @@
...
+      - uses: actions/setup-node@v3
+        with:
+          cache: yarn
+          node-version-file: .nvmrc
+      - run: yarn install --frozen-lockfile
+      - run: yarn run build
+      - name: Copy current dist to temp folder
+        run: |
+          sudo mkdir -m 777 new_dist
+          mv dist/chrome new_dist
```
faergeek, line 26: `sudo` seems unnecessary here. As well as `mkdir`.

Why not just move the whole dir like `mv dist dist.new`?

.github/workflows/update-tests.yml
```diff
@@ -0,0 +1,51 @@
...
+      # Build last tag
+      - name: Find last tag
+        run: echo ::set-output name=VALUE::$(git describe --abbrev=0 --tags `git rev-list --tags --skip=1 --max-count=1`)
```
faergeek, line 29: ::set-output is deprecated
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

test/Update.ts
```diff
@@ -0,0 +1,76 @@
...
+    );
+    await TopMenu.backButton.click();
+    return [activeAccountName, ...otherAccountNames];
+  }
+
+  it('accounts persist on update', async () => {
+    const promisifiedExec = util.promisify(exec);
+    await browser.openKeeperExtensionPage();
+    await promisifiedExec('rm -rf dist/chrome');
+    await promisifiedExec('mv new_dist/chrome dist');
```
faergeek, line 58: You can use `rename` instead, like this:

```javascript
import { rename } from 'node:fs/promises';
```

faergeek (approved):

smelnikov (approved):
I think we need more scenarios, at least 4 are defined in TestRails.
Also it would be useful to test update in all supported browsers.

PR #679, opened 2023-02-06
update balances when switching accounts

domnikov-timofei:


smelnikov (commented):
It looks like the balance will be updated in 5 seconds after switching the account.
As far as I understand the task, you need to update the balance immediately.

faergeek (approved):
I think it's better to get rid of `restartPolling`, especially because right now `restartPolling` is called from the outside in a few places. `Browser.alarms.create` call can be moved to the start of `updateCurrentAccountBalance`.

To avoid hitting concurrent requests limit (so we don't fire new requests while older requests are still in flight), it may be a good idea to also implement one of the following:

1. Set new alarm only after fetching is done. But for it to really work like `setTimeout` you'll need to call [`clear`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/alarms/clear) at the start of `updateCurrentAccountBalance`.
2. Abort previous requests before firing a new one, using [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController)

If you choose to implement one of these, make sure it doesn't break in case of failing requests.

PR #683, opened 2023-02-06
display NFTs from Puzzle Market

domnikov-timofei:


src/nfts/vendors/puzzle.ts
```diff
@@ -0,0 +1,81 @@
...
+
+const PUZZLE_MARKET_DAPP = '3PFQjjDMiZKQZdu5JqTHD7HwgSXyp9Rw9By';
+
+function puzzleDataUrl(nodeUrl: string) {
+  return new URL(`addresses/data/${PUZZLE_MARKET_DAPP}`, nodeUrl).toString();
+}
+
+function nftIdKey(id: string, postfix = '') {
+  return `nft_${id}${postfix}`;
+}
```
smelnikov, line 31: Maybe include `_` from `postfix` in string format and rename the function to `nftPropertyKey`?

src/nfts/vendors/puzzle.ts
```diff
@@ -0,0 +1,81 @@
...
+    )
+      .then(dataEntriesToRecord)
+      .then(artworksEntries => {
+        return nftIds.map((id): PuzzleNftInfo => {
+          const creator = artworksEntries[nftIdKey(id, '_issuer')];
+          const image = artworksEntries[nftIdKey(id, '_image')];
+
+          return {
+            id,
+            vendor: NftVendorId.Puzzle as NftVendorId.Puzzle,
```
smelnikov, line 59: Why is it necessary to use a type assertion here?

smelnikov (approved):

src/nfts/vendors/puzzle.ts
```diff
@@ -0,0 +1,83 @@
...
+      return [];
+    }
+
+    const nftIds = nfts.map(nft => nft.assetId);
+
+    return fetchDataEntries<DataTransactionEntryString>(
+      puzzleDataUrl(nodeUrl),
+      nftIds
+        .map(id => [nftPropertyKey(id, 'image'), nftPropertyKey(id, 'issuer')])
+        .flat()
```
faergeek, line 51: [`flatMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap) can be used instead of `map` + `flat`

faergeek (approved):

smelnikov (approved):

src/nfts/nftCard.tsx
```diff
@@ -12,26 +13,64 @@ export function NftCover({
...
-      className={clsx(styles.cover, className, isLoading && 'skeleton-glow')}
-      style={nft?.background}
-      onLoad={() => nft?.foreground && setLoading(false)}
-    />
+  useEffect(() => {
+    if (nft?.foreground) {
+      setLoading(true);
+
+      fetch(nft.foreground)
+        .then(response => (response.ok ? response : Promise.reject()))
```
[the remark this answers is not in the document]
faergeek: I'd say it's better to avoid mixing `async/await` with `.then/.catch`. But to use `async/await` I would do `Promise.resolve().then(async ...` and do everything in that function, without following `.catch/.finally`, instead using `try/catch/finally` in an async function. In general either one way or another, without mixing, except initial `Promise.resolve`, just because you can't use `async/await` inside `useEffect`.

Also, empty reject doesn't make any sense, so it makes sense to use `throw` in any case, with `async/await` or not. Because you can't throw "nothing", but you can reject with nothing.

faergeek (approved):

smelnikov (approved):

faergeek (approved):

PR #693, opened 2023-02-10
place "add account" button in the header

domnikov-timofei:


smelnikov (approved):
Maybe we can add a tooltip when hovering over the button?

src/ui/components/pages/otherAccounts.module.css
```diff
@@ -10,10 +10,17 @@
...
   top: 0;
   padding: 14px;
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
   border-bottom: 1px solid #edf0f4;
   background: #fff;
 }
 
+.header .addAccountButton {
```
faergeek, line 20: Better extract the rule defining `margin` and use it where needed than overriding existing rule when it's not needed

src/ui/components/pages/otherAccounts.module.css
```diff
@@ -55,3 +62,9 @@
 .addAccountButton:hover {
   background-color: #edf0f4;
 }
+
+.addAccountButton.addAccountButtonSmall {
```
faergeek, line 66: Same thing here. Better split into `.addAccountButton`, `.addAccountButton_text` (or `*_full`, `*_normal`, whatever) and `.addAccountButton_icon` (or `*_small`) rules to avoid overriding. In general, adding is easier to reason about than overriding.

src/ui/components/pages/otherAccounts.tsx
```diff
@@ -60,6 +61,22 @@ export function OtherAccountsPage() {
...
         <h2 className={styles.title}>{t('otherAccounts.title')}</h2>
+
+        <button
+          className={clsx(
+            styles.addAccountButton,
+            styles.addAccountButtonSmall
+          )}
+          data-testid="addAccountButton"
+          type="button"
+          onClick={() => {
```
faergeek, line 72: Better extract this function than duplicating. I doubt it would ever be different on one of these buttons, so it makes sense just to reuse this code. Doesn't matter if it's small, it's just meant to be the same.

src/ui/components/pages/otherAccounts.tsx
```diff
@@ -60,6 +61,22 @@ export function OtherAccountsPage() {
     <div className={styles.root} data-testid="otherAccountsPage">
       <header className={styles.header}>
         <h2 className={styles.title}>{t('otherAccounts.title')}</h2>
+
+        <button
+          className={clsx(
+            styles.addAccountButton,
+            styles.addAccountButtonSmall
+          )}
+          data-testid="addAccountButton"
```
faergeek, line 70: Not sure if it makes sense to set this `data-testid` on both buttons

faergeek (changes requested):

src/ui/components/pages/otherAccounts.tsx
```diff
@@ -51,6 +53,11 @@ export function OtherAccountsPage() {
     ])
   );
 
+  const addAccount = useCallback(() => {
```
faergeek, line 56: Usually there needs to be a real reason for using `useCallback` or `useMemo` instead of using it for everywhere upfront without measuring if it makes any difference :-)

faergeek (approved):

PR #1059, opened 2023-07-06
Fix attachment style

domnikov-timofei:


src/messages/transaction/common/base58.module.css
```diff
@@ -4,6 +4,7 @@
   border-radius: 2px;
   background: #f8f9fb;
   font-family: monospace;
+  word-break: break-all;
```
faergeek, line 7: `overflow-wrap: anywhere;` would be a bit nicer

faergeek (approved):

